PR #3, opened 2018-09-18
Fixes for CI

tnozicka:


.gitignore
```diff
@@ -1,20 +1 @@
-/_output
```
deads2k: pretty sure we want this and the IDE ones.

tnozicka: `_output` is not generated anymore

IDE ones don't belong there but to user's personal global .gitignore file

deads2k: > IDE ones don't belong there but to user's personal global .gitignore file

Perhaps purity-wise, but I don't see a reason to make life difficult.  Also, just saw that you updated this.  If it goes green then I'll close mine.

deads2k:
removing WIP

/lgtm

Let's get lucky.

deads2k:
@tnozicka well, I have a working verify now and a broken unit : https://github.com/openshift/cluster-openshift-apiserver-operator/pull/6

PR #4, opened 2018-09-18
Update README.md

deads2k:
kick the bot


deads2k:
@tnozicka something failed.  Looks like set up problems.

PR #7, opened 2018-09-20
simplify install path

deads2k:
Simplifies the install path for resources which require no modification.

/assign @sttts @mfojtik


deads2k:
/hold 

for upstream.

PR #8, opened 2018-09-20
update to flat updatepayload structure

juanvallejo:
Updates Dockerfile to include a `/manifests` directory in image.

cc @deads2k @mfojtik @sttts


Dockerfile
```diff
@@ -9,3 +9,6 @@ RUN cd /go/src/github.com/openshift/cluster-openshift-apiserver-operator && go b
 
 FROM centos:7
 COPY --from=0 /go/src/github.com/openshift/cluster-openshift-apiserver-operator/cluster-openshift-apiserver-operator /usr/bin/cluster-openshift-apiserver-operator
+
+COPY manifests/install /manifests
```
juanvallejo: @deads2k would we need to split the single operator.yaml file under the install folder into one file per resource? I'm assuming we'd need to name each file accordingly to ensure some resources are created before others?

deads2k: > @deads2k would we need to split the single operator.yaml file under the install folder into one file per resource? I'm assuming we'd need to name each file accordingly to ensure some resources are created before others?

split whatever is needed.  Make our `manifest` dir be in `manifest`.  Move whatever is needed.

deads2k: Also, ordering is a dangerous precedent, but if it is required, yes.

juanvallejo:
@deads2k thanks, review comments addressed

deads2k:
/lgtm

fix up the controllers too

juanvallejo:
@deads2k 
> fix up the controllers too

Opened pull here: https://github.com/openshift/cluster-openshift-controller-manager-operator/pull/6

PR #15, opened 2018-09-26
update manifest filenames to conform to expected format

juanvallejo:
Updating per https://github.com/openshift/origin/blob/master/pkg/oc/cli/admin/release/new.go#L566
Will update the cluster-openshift-controller-manager-operator as well if this update makes sense

cc @smarterclayton @deads2k


abhinavdahiya:
I think the files should be 0000_<run level>_<operator are>_<>

juanvallejo:
@abhinavdahiya @smarterclayton thanks, updated to include run level

juanvallejo:
cc @rajatchopra

smarterclayton:
Your component should be dash separated, exactly like your repository name.

juanvallejo:
@smarterclayton 
> Your component should be dash separated, exactly like your repository name.

Thanks, done

PR #16, opened 2018-09-28
 observe configuration from the cluster

deads2k:
/hold
holding for library-go

/assign @sttts @mfojtik @soltysh 

This is where my head is at for observing configuration.  We merge it in just like any other config and by having a separate control loop to watch the cluster and update it, our existing merging code works fine.  Separating the observed values from the user defined values ensures that we never let them override our changes by accident.  if they do it on purpose, it will be stopped shortly there-after.


pkg/operator/observe_config.go
```diff
@@ -0,0 +1,175 @@
...
+	if err != nil {
+		return err
+	}
+	kubeAPIServerPublicInfo, err := c.kubeClient.CoreV1().ConfigMaps(kubeAPIServerNamespaceName).Get("public-info", metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+	observedConfig := map[string]interface{}{}
+
+	if val, ok := kubeAPIServerPublicInfo.Data["projectConfig.defaultNodeSelector"]; ok {
```
soltysh: Mind adding a word of comment why we always stomp `defaultNodeSelector`?

pkg/operator/observe_config.go
```diff
@@ -0,0 +1,175 @@
...
+func (c ConfigObserver) sync() error {
+	operatorConfig, err := c.operatorConfigClient.OpenShiftAPIServerOperatorConfigs().Get("instance", metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+	kubeAPIServerPublicInfo, err := c.kubeClient.CoreV1().ConfigMaps(kubeAPIServerNamespaceName).Get("public-info", metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+	observedConfig := map[string]interface{}{}
```
soltysh: I'm missing setting this fields, the actual merge action, where is it?

pkg/operator/observe_config.go
```diff
@@ -0,0 +1,175 @@
...
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/util/wait"
+	kubeinformers "k8s.io/client-go/informers"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/util/flowcontrol"
+	"k8s.io/client-go/util/workqueue"
```
soltysh: Mind cleaning these imports, it bleeds my eyes, and I know @mfojtik will appreciate that one as well :wink:

vendor/github.com/openshift/library-go/pkg/operator/resource/resourcemerge/generic_config_merger.go
```diff
@@ -40,11 +43,10 @@ func MergeProcessConfig(specialCases map[string]MergeFunc, configYAMLs ...[]byte
 			// maybe it's just json
 			prevConfigJSON = currentConfigYAML
 		}
-		prevConfigObj, err := runtime.Decode(unstructured.UnstructuredJSONScheme, prevConfigJSON)
-		if err != nil {
+		prevConfig := map[string]interface{}{}
+		if err := json.NewDecoder(bytes.NewBuffer(prevConfigJSON)).Decode(&prevConfig); err != nil {
```
soltysh: What was wrong with previous decode into unstructured? Is this direct json encoding better/simpler/more performant?

deads2k: > What was wrong with previous decode into unstructured? Is this direct json encoding better/simpler/more performant?

Decoding into unstructured requires a GVK be specified and a map doesn't

pkg/operator/observe_config.go
deads2k: > I'm missing setting this fields, the actual merge action, where is it?

that happens when the final config is produced.

pkg/operator/observe_config.go
deads2k: > Mind adding a word of comment why we always stomp `defaultNodeSelector`?

sure.

pkg/apis/openshiftapiserver/v1alpha1/types.go
```diff
@@ -21,24 +21,28 @@ type OpenShiftAPIServerConfig struct {
...
+	// userConfig holds a sparse config that the user wants for this component.  It only needs to be the overrides from the defaults
 	// it will end up overlaying in the following order:
 	// 1. hardcoded default
 	// 2. this config
-	OpenShiftAPIServerConfig runtime.RawExtension `json:"kubeApiserverConfig" protobuf:"bytes,2,opt,name=kubeApiserverConfig"`
+	UserConfig runtime.RawExtension `json:"userConfig"`
+
+	// observedConfig holds a sparse config that controller has observed from the cluster state.  It exists in spec because
+	// it causes action for the operator
+	ObservedConfig runtime.RawExtension `json:"observedConfig"`
```
deads2k: @smarterclayton this field

deads2k: The `observedConfig` is an input to the level that determines an overall config.  The overall config (logically the merging of default, user, and observed) is a level (reflected by spec.generation) that the operator will try to reconcile.  Once it has successfully reconciled the level, status.generation will indicate that the level has been achieved.  Since it is an input the level, it should live in the spec, not the status.

deads2k: > If we decide this was a terrible idea in the future, can we change it? I.e. can we classify these APIs in a way that gives us more flexibility than we might traditionally have to change their internal workings?

Yes.  Since they will be immediately stomped by a controller, the "nothing you ever set here had an effect" rule applies to changes.

tnozicka: API wise, to me that name sounds ambiguous - in the controller context *observed* means something that the controller saw and acted upon. I don't have a good fitting name it though.

> I can see the reasoning for this, and it's probably the best way to manage complex config opaquely.

@smarterclayton 

Is there a reason why we need to cache the value? Currently it are the etcd urls - getting them from endpoints informer is as cheep as it gets. Caching the field on the object itself leads to having 2 controllers on the same object just to fill in this field, introducing complexity, loosing atomicity, RV collisions, likely broken first rollout until the other loop catches up, ...

I would imagine this config detected from cluster state is determined in a regular sync loop from informer caches, merged with what we have in the spec and not stored on the operator config object.

I disagree with the generation implications presented here. Generation is good for letting user know which version the controller saw and acted upon. We shouldn't trigger based on generation but on actual!=desired. Like when in sync loop you merge UserConfig and etcd values you read from endpoints and the result doesn't match what the running apiserver is configured with. If we don't need it for triggering, as generation is not what we should be triggering on, and it is not something anyone should set and we don't need to cache it, I don't see a reason to have this field and multiple controllers coupled to it.

/cc @mfojtik

PR #19, opened 2018-10-02
only install on master nodes

deads2k:


bindata/v3.11.0/openshift-apiserver/ds.yaml
```diff
@@ -57,6 +57,5 @@ spec:
       - name: serving-cert
         secret:
           secretName: serving-cert
-
```
mfojtik: hrm. what are this 3 empty lines? ;-)

PR #35, opened 2018-10-10
fix deployment env

juanvallejo:
cc @deads2k


sttts:
Lgtm.

Can and should we derive this from the image reference file during build?

juanvallejo:
> Can and should we derive this from the image reference file during build?

cc @deads2k

PR #36, opened 2018-10-15
Abstract ensureOperatorConfigExists to use Dynamic client

damemi:
Plan is to pull out the operator-specific approach into a method that can hopefully just be added to library-go and used in multiple operators.

@deads2k could you review my approach here and see if anything stands out to you? If this looks good I'll move ahead trying to get this into library-go and then update this PR to use that instead


pkg/operator/starter.go
```diff
@@ -111,9 +130,17 @@ func ensureOperatorConfigExists(client operatorsv1alpha1client.OpenShiftAPIServe
...
 
 	// If ImagePullSpec changed, update the existing config instance
-	if existing.Spec.ImagePullSpec != requiredOperatorConfig.Spec.ImagePullSpec {
-		existing.Spec.ImagePullSpec = requiredOperatorConfig.Spec.ImagePullSpec
-		if _, err := client.OpenShiftAPIServerOperatorConfigs().Update(existing); err != nil {
+	existingSpec, _, err := unstructured.NestedMap(existing.Object, "Spec")
+	if err != nil {
+		panic(err)
+	}
+	if existingSpec["ImagePullSpec"] != spec["ImagePullSpec"] {
```
deads2k: There is a get nested string I think

pkg/operator/starter.go
```diff
@@ -75,29 +82,41 @@ func RunOperator(clientConfig *rest.Config, stopCh <-chan struct{}) error {
 	return fmt.Errorf("stopped")
 }
 
-func ensureOperatorConfigExists(client operatorsv1alpha1client.OpenShiftAPIServerOperatorConfigsGetter, path string) {
-	v1alpha1Scheme := runtime.NewScheme()
-	v1alpha1.Install(v1alpha1Scheme)
-	v1alpha1Codecs := serializer.NewCodecFactory(v1alpha1Scheme)
+func ensureOperatorConfigExists(client dynamic.Interface, path string) {
+	// TODO pull out MustAsset from this repo
```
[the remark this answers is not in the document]
damemi: Looking into this I'm confused on how we would do that. `MustAsset` appears to be generated by `go-bindata`, so if the bindata is generated and used specifically by each operator is there a way to generalize that? Maybe not use bindata at all?

pkg/operator/starter.go
```diff
@@ -75,29 +82,41 @@ func RunOperator(clientConfig *rest.Config, stopCh <-chan struct{}) error {
...
+		}
+		spec["ImagePullSpec"] = imagePullSpecFromEnv
+		err = unstructured.SetNestedMap(requiredOperatorConfig.Object, spec, "Spec")
+		if err != nil {
+			panic(err)
+		}
 	}
 
-	existing, err := client.OpenShiftAPIServerOperatorConfigs().Get(requiredOperatorConfig.Name, metav1.GetOptions{})
+	existing, err := client.Resource(schema.GroupVersionResource{Group: v1alpha1.GroupName, Version: "v1alpha1", Resource: ""}).Get(requiredOperatorConfig.GetName(), metav1.GetOptions{})
```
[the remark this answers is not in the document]
damemi: What is the Resource for this? `OpenshiftAPIServerOperatorConfig`?

pkg/operator/starter.go
```diff
@@ -43,7 +45,10 @@ func RunOperator(clientConfig *rest.Config, stopCh <-chan struct{}) error {
 	if err != nil {
 		return err
 	}
-	ensureOperatorConfigExists(operatorConfigClient.OpenshiftapiserverV1alpha1(), "v3.11.0/openshift-apiserver/operator-config.yaml")
+
+	path := "v3.11.0/openshift-apiserver/operator-config.yaml"
```
deads2k: not need to assign here.

pkg/operator/starter.go
```diff
@@ -43,7 +45,10 @@ func RunOperator(clientConfig *rest.Config, stopCh <-chan struct{}) error {
...
+	operatorConfigBytes := v311_00_assets.MustAsset(path)
```
deads2k: nit: inline in the function.  We never use this again.

damemi:
Switched this to use the library-go function, and will squash once https://github.com/openshift/library-go/pull/53 merges

deads2k:
do a real vendor

PR #37, opened 2018-10-16
 add ClusterOperator status

deads2k:
uses the same library-go logic as controller-manager


deads2k:
congruent to already merged pull

PR #39, opened 2018-10-23
[WIP] - Use common build machinery

tnozicka:
Remove hack/. Move to shared build-machinery.

/hold
(need to properly vendor when https://github.com/openshift/library-go/pull/55/ merges)

/cc @deads2k @sttts @openshift/sig-master


tnozicka:
@soltysh could you point to particular places not obvious to you that should have a comment explaning them? Is that about the 2 `$(call ...)` that could use it's argument names being spelled out in a comment?

I fear of comments getting outdated, so I'd appreciate a pointer where it would be most useful.

soltysh:
@tnozicka yes the `$(call ...)` operations are not ringing the bell for me so I'd appreciate any info on it.

tnozicka:
Updated to current library-go PR state and added comments

tnozicka:
/test e2e-aws
(flake - RequestLimitExceeded)

tnozicka:
now with proper bump bringing the library-go

PR #40, opened 2018-10-23
Bump deps

juanvallejo:
Bumps:
- client-go
- openshift/api
- library-go
- spf13/pflag

client-go and openshift/api bumps are needed for https://github.com/openshift/cluster-openshift-apiserver-operator/pull/41

cc @deads2k


deads2k:
squash all bumps and self tag  lgtm

PR #45, opened 2018-11-06
pick up correct source of etcd connection info

deads2k:
This information actually originates in the kube-system (or etcd) namespace.

/assign @sanchezl


deads2k:
other one was lgtmd

PR #54, opened 2018-11-19
add additional imageconfig api field handling

bparees:


mfojtik:
@bparees bump?

bparees:
@mfojtik this can't proceed until https://github.com/openshift/api/pull/126 merges, and https://github.com/openshift/api/pull/126 can't proceed until https://github.com/openshift/origin/pull/21490 merges and https://github.com/openshift/origin/pull/21490 can't proceed until you guys are done w/ the rebase, as I understand it, because you don't want more bumps in origin.

https://github.com/openshift/origin/pull/21490 is also blocked by the fact that the SCSS team broke bumps to origin (there is another PR in flight to fix that).

bparees:
@deads2k this is ready for review now.  If you want a true example of the patterns other people will need to follow, I think this is it.

/assign @deads2k

pkg/operator/configobservation/images/observe_images.go
```diff
@@ -1,42 +1,111 @@
...
-		return prevObservedConfig, errs
+		return prevObservedConfig, append(errs, err)
 	}
+
 	internalRegistryHostName := configImage.Status.InternalRegistryHostname
 	if len(internalRegistryHostName) > 0 {
-		unstructured.SetNestedField(observedConfig, internalRegistryHostName, internalRegistryHostnamePath...)
+		err = unstructured.SetNestedField(observedConfig, internalRegistryHostName, internalRegistryHostnamePath...)
+		if err != nil {
+			return observedConfig, append(errs, err)
```
bparees: @deads2k @sanchezl it's not obvious to me whether errors at this point in the process should return prevObservedConfig or observedConfig.  Returning observedConfig here means we're potentially throwing away existing config.  Returning prevObservedConfig means potentially returning invalid/stale information.

(same applies to all the other places where we catch an error and return observedConfig below)

bparees:
@deads2k bump

pkg/operator/configobservation/images/observe_images.go
```diff
@@ -1,42 +1,111 @@
...
-		unstructured.SetNestedField(prevObservedConfig, currentInternalRegistryHostname, internalRegistryHostnamePath...)
+		err := unstructured.SetNestedField(prevObservedConfig, currentInternalRegistryHostname, internalRegistryHostnamePath...)
+		if err != nil {
+			return prevObservedConfig, append(errs, err)
+		}
+	}
+
+	externalRegistryHostnamePath := []string{"imagePolicyConfig", "externalRegistryHostnames"}
+	o, _, err := unstructured.NestedSlice(existingConfig, externalRegistryHostnamePath...)
+	err = unstructured.SetNestedSlice(prevObservedConfig, o, externalRegistryHostnamePath...)
```
deads2k: place in len block for consistency?

pkg/operator/configobservation/images/observe_images.go
```diff
@@ -1,42 +1,111 @@
...
+	if err != nil {
+		return prevObservedConfig, append(errs, err)
+	}
+
+	allowedRegistriesForImportPath := []string{"imagePolicyConfig", "allowedRegistriesForImport"}
+	o, _, err = unstructured.NestedSlice(existingConfig, allowedRegistriesForImportPath...)
+	err = unstructured.SetNestedSlice(prevObservedConfig, o, allowedRegistriesForImportPath...)
```
deads2k: if block?

deads2k:
reading through, I would have separated the methods based on my destination fields.  That clarifies what to do with previous config versus observed config.

deads2k:
The encoding block looks weird.  You may want confirm that it is encoding as you expect.

pkg/operator/configobservation/images/observe_images.go
bparees: consistency with what?  internalRegistryHostname is a string, this is a slice, there are different expectations for what empty means.

pkg/operator/configobservation/images/observe_images.go
bparees: this (as above) was intentional... if the existing config has an empty slice for this value, i want to set the prevObservedConfig to be that empty slice, so that the config gets updated with that empty slice value instead of whatever slice content may be there currently.

sanchezl:
> reading through, I would have separated the methods based on my destination fields. That clarifies what to do with previous config versus observed config.

While I agree with that, how does that fit in with the previous guidance on internal vs external names? 
For example in this pull request, `ObserveInternalRegistryHostname` is renamed to `ObserveImageConfig`.

bparees:
> The encoding block looks weird. You may want confirm that it is encoding as you expect.

it looked right to me when i checked it, plus the unit test is presumably confirming it, no?

bparees:
> reading through, I would have separated the methods based on my destination fields.

are you suggesting a separate Observer per field?  That seems painful.

bparees:
@deads2k updated to use an observer func per field.  retaining the other behavior per our discussion

pkg/operator/configobservation/images/observe_images.go
```diff
@@ -1,42 +1,163 @@
...
+	"encoding/json"
+
 	"github.com/golang/glog"
 
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 
-	"github.com/openshift/cluster-openshift-apiserver-operator/pkg/operator/configobservation"
 	"github.com/openshift/library-go/pkg/operator/configobserver"
+
```
deads2k: nit: unusual newline

pkg/operator/configobservation/images/observe_images.go
```diff
@@ -1,42 +1,163 @@
...
 		return prevObservedConfig, errs
 	}
 
+	// now gather the cluster config and turn it into the observed config
 	observedConfig := map[string]interface{}{}
 	configImage, err := listers.ImageConfigLister.Get("cluster")
 	if errors.IsNotFound(err) {
 		glog.Warningf("image.config.openshift.io/cluster: not found")
-		return observedConfig, errs
+		return observedConfig, append(errs, err)
```
deads2k: this error shouldn't be an error that gets returned because it doesn't indicate a real failure.  there's just nothing to report

pkg/operator/configobservation/images/observe_images.go
```diff
@@ -1,42 +1,163 @@
...
 	}
 	if err != nil {
-		return prevObservedConfig, errs
+		return prevObservedConfig, append(errs, err)
```
deads2k: this one is correct

pkg/operator/configobservation/images/observe_images.go
```diff
@@ -1,42 +1,163 @@
...
+
+func ObserveExternalRegistryHostnames(genericListers configobserver.Listers, existingConfig map[string]interface{}) (map[string]interface{}, []error) {
+	listers := genericListers.(configobservation.Listers)
+	var errs []error
+	prevObservedConfig := map[string]interface{}{}
+
+	// first observe all the existing config values so that if we get any errors
+	// we can at least return those.
+	externalRegistryHostnamePath := []string{"imagePolicyConfig", "externalRegistryHostnames"}
+	o, _, err := unstructured.NestedSlice(existingConfig, externalRegistryHostnamePath...)
```
deads2k: accidentally ignoring this error?

pkg/operator/configobservation/images/observe_images.go
```diff
@@ -1,42 +1,163 @@
...
+		glog.Warning("images.config.openshift.io not synced")
+		return prevObservedConfig, errs
+	}
+
+	// now gather the cluster config and turn it into the observed config
+	observedConfig := map[string]interface{}{}
+	configImage, err := listers.ImageConfigLister.Get("cluster")
+	if errors.IsNotFound(err) {
+		glog.Warningf("image.config.openshift.io/cluster: not found")
+		return observedConfig, append(errs, err)
```
deads2k: same comment about not an error.

pkg/operator/configobservation/images/observe_images.go
```diff
@@ -1,42 +1,163 @@
...
```
deads2k: not an error.

deads2k:
a few straight-forward. lgtm otherwise.  You want to add an OWNERS file in this directory for yourself?

/approve

pkg/operator/configobservation/images/observe_images.go
bparees: I separate imports from the same repo from imports from other repos.  Call it a personal quirk but i find it makes it easier to scan what's being imported from where.

I basically group imports into:

```
system

random thirdparties

k8s

openshift(other repos)

current repo
```

pkg/operator/configobservation/images/observe_images.go
bparees: will fix

pkg/operator/configobservation/images/observe_images.go
bparees: yeah, will fix

pkg/operator/configobservation/images/observe_images.go
bparees: will fix

pkg/operator/configobservation/images/observe_images.go
bparees: will fix

PR #62, opened 2018-11-26
force rollout on input change managed by a different controller

deads2k:
This watches for changes to a configmap and can cause a rollout of the deployment

/assign @mfojtik @sttts

### DIFF
--- a/pkg/operator/detectinputchange/as_configmap.go
+++ b/pkg/operator/detectinputchange/as_configmap.go
@@ -1,0 +1,101 @@
+package detectinputchange
+
+import (
+	"encoding/base64"
+	"encoding/json"
+	"fmt"
+	"hash/fnv"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/kubernetes"
+)
+
+// GetConfigMapHash returns a hash of the configmap data
+func GetConfigMapHash(obj *corev1.ConfigMap) (string, error) {
+	hasher := fnv.New32()
+	if err := json.NewEncoder(hasher).Encode(obj.Data); err != nil {
+		return "", err
+	}
+	return base64.URLEncoding.EncodeToString(hasher.Sum(nil)), nil
+}
+
+// GetSecretHash returns a hash of the secret data
+func GetSecretHash(obj *corev1.Secret) (string, error) {
+	hasher := fnv.New32()
+	if err := json.NewEncoder(hasher).Encode(obj.Data); err != nil {
+		return "", err
+	}
+	return base64.URLEncoding.EncodeToString(hasher.Sum(nil)), nil
+}
+
+// MultipleObjectHashStringMap returns a map of key/hash pairs suitable for merging into a configmap
+func MultipleObjectHashStringMap(objs ...runtime.Object) (map[string]string, error) {
+	ret := map[string]string{}
+
+	for _, obj := range objs {
+		switch t := obj.(type) {
+		case *corev1.ConfigMap:
+			hash, err := GetConfigMapHash(t)
+			if err != nil {
+				return nil, err
+			}
+			// this string coercion is lossy, but it should be fairly controlled and must be an allowed name
+			ret[mapKeyFor("configmap", t.Namespace, t.Name)] = hash
+
+		case *corev1.Secret:
+			hash, err := GetSecretHash(t)
+			if err != nil {
+				return nil, err
+			}
+			// this string coercion is lossy, but it should be fairly controlled and must be an allowed name
+			ret[mapKeyFor("secret", t.Namespace, t.Name)] = hash
+
+		default:
+			return nil, fmt.Errorf("%T is not handled", t)
+		}
+	}
+
+	return ret, nil
+}
+
+func mapKeyFor(resource, namespace, name string) string {
+	return fmt.Sprintf("%s.%s.%s", namespace, name, resource)
+}
+
+// ObjectReference can be used to reference a particular resource.  Not all group resources are respected by all methods.
+type ObjectReference struct {
+	Resource  schema.GroupResource
+	Namespace string
+	Name      string
+}
+
+// MultipleObjectHashStringMapForObjectReferences returns a map of key/hash pairs suitable for merging into a configmap
+func MultipleObjectHashStringMapForObjectReferences(client kubernetes.Interface, objRefs ...ObjectReference) (map[string]string, error) {
+	objs := []runtime.Object{}
+
+	for _, objRef := range objRefs {
+		switch objRef.Resource {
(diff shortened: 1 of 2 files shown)


deads2k:
@enj

deads2k:
@mfojtik now rebased to make it smaller.  does this make sense?  It was @bparees idea to include the hashes in an already tracked object.  I think it works pretty well.

deads2k:
```
   stderr:
    panic: boringcrypto: not in FIPS mode
```

that's fun

/retest

deads2k:
@mfojtik now green.  We need something like this to do cert rotation easily.

pkg/operator/detectinputchange/as_configmap.go
```diff
@@ -0,0 +1,115 @@
...
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/runtime/serializer"
+	"k8s.io/client-go/kubernetes"
+)
+
+func GetConfigMapHash(obj *corev1.ConfigMap) (string, error) {
```
sttts: godocs

pkg/operator/detectinputchange/as_configmap.go
```diff
@@ -0,0 +1,115 @@
...
+func GetSecretHash(obj *corev1.Secret) (string, error) {
+	hasher := fnv.New32()
+	_, err := hasher.Write([]byte(WriteSecretV1OrDie(obj)))
+	if err != nil {
+		return "", err
+	}
+	return base64.URLEncoding.EncodeToString(hasher.Sum(nil)), nil
+}
+
+func GetStringMapDataHashesForObjects(objs ...runtime.Object) (map[string]string, error) {
```
sttts: MultipleObjectHashStringMap

pkg/operator/detectinputchange/as_configmap.go
```diff
@@ -0,0 +1,115 @@
...
+		default:
+			return nil, fmt.Errorf("%v is not handled", objRef.Resource)
+		}
+	}
+
+	return GetStringMapDataHashesForObjects(objs...)
+}
+
+// TODO move to library-go
+func WriteConfigMapV1OrDie(obj *corev1.ConfigMap) string {
```
sttts: write?

pkg/operator/detectinputchange/as_configmap.go
mfojtik: also looks like something that should live in library-go

pkg/operator/detectinputchange/as_configmap.go
```diff
@@ -0,0 +1,115 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/runtime/serializer"
+	"k8s.io/client-go/kubernetes"
+)
+
+func GetConfigMapHash(obj *corev1.ConfigMap) (string, error) {
+	hasher := fnv.New32()
+	_, err := hasher.Write([]byte(WriteConfigMapV1OrDie(obj)))
```
mfojtik: if _, err := hasher...

pkg/operator/detectinputchange/as_configmap.go
```diff
@@ -0,0 +1,115 @@
...
+	if err != nil {
+		return "", err
+	}
+	return base64.URLEncoding.EncodeToString(hasher.Sum(nil)), nil
+}
+
+func GetSecretHash(obj *corev1.Secret) (string, error) {
+	hasher := fnv.New32()
+	_, err := hasher.Write([]byte(WriteSecretV1OrDie(obj)))
```
mfojtik: nit: if _, err := hasher...

pkg/operator/detectinputchange/as_configmap.go
```diff
@@ -0,0 +1,115 @@
...
+
+	for _, obj := range objs {
+		switch t := obj.(type) {
+		case *corev1.ConfigMap:
+			hash, err := GetConfigMapHash(t)
+			if err != nil {
+				return nil, err
+			}
+			// this string coercion is lossy, but it should be fairly controlled and must be an allowed name
+			ret[fmt.Sprintf("%s.%s.%s", t.Namespace, t.Name, "configmap")] = hash
```
mfojtik: nit: move `fmt.Sprintf("%s.%s.%s", t.Namespace, t.Name, "configmap")` to function

pkg/operator/detectinputchange/as_configmap.go
```diff
@@ -0,0 +1,115 @@
...
+
+		case *corev1.Secret:
+			hash, err := GetSecretHash(t)
+			if err != nil {
+				return nil, err
+			}
+			// this string coercion is lossy, but it should be fairly controlled and must be an allowed name
+			ret[fmt.Sprintf("%s.%s.%s", t.Namespace, t.Name, "secret")] = hash
```
mfojtik: nit: replace with function

pkg/operator/detectinputchange/as_configmap.go
```diff
@@ -0,0 +1,115 @@
...
+		default:
+			return nil, fmt.Errorf("%T is not handled", t)
+		}
+	}
+
+	return ret, nil
+}
+
+// ObjectReference can be used to reference a particular resource.  Not all group resources are respected by all methods.
+type ObjectReference struct {
```
mfojtik: why not use v1.ObjectReference ?

pkg/operator/detectinputchange/as_configmap.go
```diff
@@ -0,0 +1,115 @@
...
+	Resource  schema.GroupResource
+	Namespace string
+	Name      string
+}
+
+func GetStringMapDataHashesForObjectReferences(client kubernetes.Interface, objRefs ...ObjectReference) (map[string]string, error) {
+	objs := []runtime.Object{}
+
+	for _, objRef := range objRefs {
+		switch objRef.Resource {
```
mfojtik: nit: objRef.Resource.Resource ? then you don't need to make the schema

pkg/operator/detectinputchange/as_configmap.go
deads2k: > why not use v1.ObjectReference ?

Too many fields that don't make sense.  We've moved away from using it in APIs

deads2k:
updated

bparees:
How does this compare/contrast with resourceapply.SyncConfigMap ?

I'm trying to solve the use case of a configmap in a different namespace (openshift-config) that i need to copy into the api-server namespace on change, and then rollout the apiserver.

I suppose the two pieces can just be combined, ie use the syncconfigmap to copy it, and this new code to trigger the rollout when the local CM gets updated by the sync?

deads2k:
> I suppose the two pieces can just be combined, ie use the syncconfigmap to copy it, and this new code to trigger the rollout when the local CM gets updated by the sync?

Yes, they are separate,but can be combined

PR #64, opened 2018-11-28
Fix handling of empty lists of registries

Miciah:
* `pkg/operator/configobservation/images/observe_images.go` (`ObserveExternalRegistryHostnames`, `ObserveAllowedRegistriesForImport`): Ignore empty `externalRegistryHostnames` or `allowedRegistriesForImport` values in old observed config, and omit them from new observed config.
* `pkg/operator/configobservation/images/observe_images_test.go`: Expect to get null for empty `allowedRegistriesForImport`.

These changes follow up #54 to avoid spuriously adding and removing `imagePolicyConfig` with null values, as observed in the operator's log output:
```
I1128 06:10:44.254039       1 config_observer_controller.go:115] writing updated observedConfig: {

A: }

B: "imagePolicyConfig":{"allowedRegistriesForImport":null,"externalRegistryHostnames":null}}

I1128 06:10:44.264287       1 config_observer_controller.go:115] writing updated observedConfig: {

A: "imagePolicyConfig":{"allowedRegistriesForImport":null,"externalRegistryHostnames":null}}

B: }
```
The same spurious adding and removing repeats while the operator runs.  I also observed the following in the openshiftapiserveroperatorconfig: 
```
status:
  conditions:
  - lastTransitionTime: null
    message: |-
      .imagePolicyConfig.allowedRegistriesForImport accessor error: <nil> is of the type <nil>, expected []interface{}
      .imagePolicyConfig.externalRegistryHostnames accessor error: <nil> is of the type <nil>, expected []interface{}
    reason: ConfigObservationError
    status: "True"
    type: ConfigObservationFailing
```

@bparees, please take a look.  I may be missing some subtlety in how null values and empty lists are handled in the API, JSON, or gRPC, but I figure the log output and status indicate that _something_ is amiss in the current logic.

### DIFF
--- a/pkg/operator/configobservation/images/observe_images.go
+++ b/pkg/operator/configobservation/images/observe_images.go
@@ -72,9 +72,11 @@ func ObserveExternalRegistryHostnames(genericListers configobserver.Listers, exi
 	if err != nil {
 		return prevObservedConfig, append(errs, err)
 	}
-	err = unstructured.SetNestedSlice(prevObservedConfig, o, externalRegistryHostnamePath...)
-	if err != nil {
-		return prevObservedConfig, append(errs, err)
+	if len(o) > 0 {
+		err = unstructured.SetNestedSlice(prevObservedConfig, o, externalRegistryHostnamePath...)
+		if err != nil {
+			return prevObservedConfig, append(errs, err)
+		}
 	}
 
 	if !listers.ImageConfigSynced() {
@@ -98,13 +100,15 @@ func ObserveExternalRegistryHostnames(genericListers configobserver.Listers, exi
 	externalRegistryHostnames := configImage.Spec.ExternalRegistryHostnames
 	externalRegistryHostnames = append(externalRegistryHostnames, configImage.Status.ExternalRegistryHostnames...)
 
-	hostnames, err := Convert(externalRegistryHostnames)
-	if err != nil {
-		return prevObservedConfig, append(errs, err)
-	}
-	err = unstructured.SetNestedField(observedConfig, hostnames, externalRegistryHostnamePath...)
-	if err != nil {
-		return prevObservedConfig, append(errs, err)
+	if len(externalRegistryHostnames) > 0 {
+		hostnames, err := Convert(externalRegistryHostnames)
+		if err != nil {
+			return prevObservedConfig, append(errs, err)
+		}
+		err = unstructured.SetNestedField(observedConfig, hostnames, externalRegistryHostnamePath...)
+		if err != nil {
+			return prevObservedConfig, append(errs, err)
+		}
 	}
 
 	return observedConfig, errs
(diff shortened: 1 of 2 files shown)


pkg/operator/configobservation/images/observe_images.go
```diff
@@ -98,13 +100,15 @@ func ObserveExternalRegistryHostnames(genericListers configobserver.Listers, exi
...
 	externalRegistryHostnames = append(externalRegistryHostnames, configImage.Status.ExternalRegistryHostnames...)
 
-	hostnames, err := Convert(externalRegistryHostnames)
-	if err != nil {
-		return prevObservedConfig, append(errs, err)
-	}
-	err = unstructured.SetNestedField(observedConfig, hostnames, externalRegistryHostnamePath...)
-	if err != nil {
-		return prevObservedConfig, append(errs, err)
+	if len(externalRegistryHostnames) > 0 {
```
bparees: i don't think this is correct.  We have to be able to blank out the value if all the externalRegistryHostnames are removed.

pkg/operator/configobservation/images/observe_images.go
```diff
@@ -122,9 +126,11 @@ func ObserveAllowedRegistriesForImport(genericListers configobserver.Listers, ex
 	if err != nil {
 		return prevObservedConfig, append(errs, err)
 	}
-	err = unstructured.SetNestedSlice(prevObservedConfig, o, allowedRegistriesForImportPath...)
-	if err != nil {
-		return prevObservedConfig, append(errs, err)
+	if len(o) > 0 {
```
bparees: same here, we need to be able to blank out this value if all the allowedregistries are removed.

Miciah:
I tested with `oc patch images.config.openshift.io/cluster --type=merge --patch='{"spec":{"allowedRegistriesForImport":null}}'` and `oc patch images.config.openshift.io/cluster --type=merge --patch='{"spec":{"allowedRegistriesForImport":[]}}'` and similar for `externalRegistryHostnames` and saw the expected output, e.g.:
```
I1128 07:43:59.145080       1 config_observer_controller.go:115] writing updated observedConfig: {

A: "imagePolicyConfig":{"allowedRegistriesForImport":[{"domainName":"b"}]}}

B: }

```
Is that what you mean? I'll try to get a cluster running so I can do more tests.

bparees:
I wouldn't call that the expected output...if you set the array to nil or empty, the config should have nil or empty, not the old value.

pkg/operator/configobservation/images/observe_images.go
deads2k: > i don't think this is correct. We have to be able to blank out the value if all the externalRegistryHostnames are removed.

`observedConfig` starts out as empty (missing entirely).  Wouldn't this code just keep that (value is entirely missing) when the list only has zero entries?

bparees: hm. I had a misunderstanding of what is done w/ the returned values.

bparees:
sorry i think i misinterpetted your output

PR #65, opened 2018-11-28
Quickfix: increase pod readiness timeout

sttts:


mfojtik:
/lgtm

@sttts can we have issue to investigate why is this needed?

sttts:
Here is a log with infinite readiness failureThreshold: https://gist.github.com/sttts/044eeb31338731f9c639d90bf6759980

deads2k:
> Works for me (on libvirt).

alright.  I'll assume CI is borked and this will unstick us.  @sttts keep an eye out to see if this helps or destroys us.

stlaz:
Actually, it seems that it makes the pod survive the first hour, but after that, it's crash-looping again.

sttts:
> Actually, it seems that it makes the pod survive the first hour, but after that, it's crash-looping again.

Citing @stlaz from Slack: 

> ah, damn, I'm on a wrong VM -.-

PR #67, opened 2018-11-28
Updates library-go to write the correct status.

deads2k:
/hold

holding for real bump


deads2k:
simple bump and a minor update.  tagging.

PR #72, opened 2018-11-29
bump library go to get events and fix cluster status

mfojtik:


deads2k:
how does this different from https://github.com/openshift/cluster-openshift-apiserver-operator/pull/67 ?

PR #73, opened 2018-11-29
make daemonset pods node critical priority

mfojtik:
/cc @deads2k 
/cc @aveshagarwal

### DIFF
--- a/bindata/v3.11.0/openshift-apiserver/ds.yaml
+++ b/bindata/v3.11.0/openshift-apiserver/ds.yaml
@@ -21,6 +21,7 @@ spec:
         apiserver: "true"
     spec:
       serviceAccountName: openshift-apiserver-sa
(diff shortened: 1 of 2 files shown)


bindata/v3.11.0/openshift-apiserver/ds.yaml
```diff
@@ -21,6 +21,7 @@ spec:
         apiserver: "true"
     spec:
       serviceAccountName: openshift-apiserver-sa
+      priorityClassName: system-node-critical
```
aveshagarwal: My guess is that operator in itself could be system-cluster-priority.

aveshagarwal: Nevermind that is actual pod.

aveshagarwal:
@mfojtik needs to assign priority to operator pods too.

PR #74, opened 2018-11-30
operator: wait for spec.observedConfig to be fill before installing

sttts:


deads2k:
I don't think we should do this.  This is a symptom of our need to get our endpoint removed before we stop responding to connections.

/hold

deads2k:
> I don't think we should do this. This is a symptom of our need to get our endpoint removed before we stop responding to connections.
> 
> /hold

do we have a list of more specific items that need to be completed before we start?

deads2k:
> do we have a list of more specific items that need to be completed before we start?

@sttts I'll take it an an IOU along with an event.

/hold cancel

/retest

PR #84, opened 2018-12-05
add memory request to openshift-apiserver and operator pods

sjenning:
One in a series of PRs to set memory requests for all control plane and high memory use components.

We are encountering master instability (https://github.com/openshift/installer/pull/408 https://github.com/openshift/installer/pull/785) as the number of components increase because many components run `BestEffort` giving the scheduler no information on pod resource usage.

On a steady state empty cluster, `openshift-apiserver` uses 179Mi.

{pod_name="apiserver-t5v2d"} | 179249152

Because it is a control plane component, I'm giving it a buffer and setting to `200Mi`

@deads2k @sttts  @derekwaynecarr


sjenning:
/hold
I did this against the operator, not the actually pod

sjenning:
fixed
/hold cancel

PR #85, opened 2018-12-06
add clusteroperator resource to CVO payload

deads2k:
Adding this resource should give the CVO visibility into whether our component is working or not.

/assign @sttts


sttts:
/lgtm
/retest

Looks like we miss something? All these PRs are red.

PR #87, opened 2018-12-07
consider api service availability when reporting status

sanchezl:


pkg/operator/workload_controller_openshiftapiserver_v311_00.go
```diff
@@ -89,17 +90,26 @@ func syncOpenShiftAPIServer_v311_00_to_latest(c OpenShiftAPIServerOperator, orig
 	}
 
 	// manage status
-	if actualDaemonSet.Status.NumberAvailable > 0 {
+	var availabilityErrors []string
+	if actualDaemonSet.Status.NumberAvailable == 0 {
+		availabilityErrors = append(availabilityErrors, fmt.Sprintf("daemonset/%v: no daemon pods available on any node.", actualDaemonSet.Name))
+	}
+	for _, currResult := range directResourceResults {
+		if currResult.Result == nil {
```
deads2k: result == nil isn't the same as an error

deads2k: > result == nil isn't the same as an error

and this sets failing.

pkg/operator/workload_controller_openshiftapiserver_v311_00.go
```diff
@@ -89,17 +90,26 @@ func syncOpenShiftAPIServer_v311_00_to_latest(c OpenShiftAPIServerOperator, orig
 	}
 
 	// manage status
-	if actualDaemonSet.Status.NumberAvailable > 0 {
+	var availabilityErrors []string
+	if actualDaemonSet.Status.NumberAvailable == 0 {
+		availabilityErrors = append(availabilityErrors, fmt.Sprintf("daemonset/%v: no daemon pods available on any node.", actualDaemonSet.Name))
+	}
+	for _, currResult := range directResourceResults {
```
deads2k: delete this entire block

deads2k: you need to make sure all of these are Conditions["Available"] == true https://github.com/openshift/cluster-openshift-apiserver-operator/blob/master/pkg/operator/workload_controller_openshiftapiserver_v311_00.go#L214-L227

pkg/operator/workload_controller_openshiftapiserver_v311_00.go
```diff
@@ -81,25 +82,51 @@ func syncOpenShiftAPIServer_v311_00_to_latest(c OpenShiftAPIServerOperator, orig
...
 		if err != nil {
 			errors = append(errors, fmt.Errorf("%q: %v", "apiservices", err))
 		}
 	}
 
 	// manage status
-	if actualDaemonSet.Status.NumberAvailable > 0 {
+	var availableConditionReason string
+	var availableConditionMessages []string
+	if actualDaemonSet != nil && actualDaemonSet.Status.NumberAvailable == 0 {
```
deads2k: Seems like this should be an OR

pkg/operator/workload_controller_openshiftapiserver_v311_00.go
```diff
@@ -81,25 +82,51 @@ func syncOpenShiftAPIServer_v311_00_to_latest(c OpenShiftAPIServerOperator, orig
...
+		availableConditionReason = "NoAPIServerNode"
+		availableConditionMessages = append(availableConditionMessages, "no openshift-apiserver daemon pods available on any node.")
+	}
+	for _, apiService := range actualAPIServices {
```
deads2k: check all these first because the last set wins and the daemonset being missing is more important to report.

pkg/operator/workload_controller_openshiftapiserver_v311_00.go
```diff
@@ -81,25 +82,51 @@ func syncOpenShiftAPIServer_v311_00_to_latest(c OpenShiftAPIServerOperator, orig
...
+		for _, condition := range apiService.Status.Conditions {
+			if condition.Type == apiregistrationv1.Available {
+				if condition.Status == apiregistrationv1.ConditionFalse {
+					availableConditionReason = "APIServiceNotAvailable"
+					availableConditionMessages = append(availableConditionMessages, fmt.Sprintf("API service %v is not available.", apiService.Name))
```
deads2k: include the apiservice message in our message.

pkg/operator/workload_controller_openshiftapiserver_v311_00.go
```diff
@@ -81,25 +82,51 @@ func syncOpenShiftAPIServer_v311_00_to_latest(c OpenShiftAPIServerOperator, orig
...
 			errors = append(errors, fmt.Errorf("%q: %v", "apiservices", err))
 		}
 	}
 
 	// manage status
-	if actualDaemonSet.Status.NumberAvailable > 0 {
+	var availableConditionReason string
+	var availableConditionMessages []string
+	if actualDaemonSet != nil && actualDaemonSet.Status.NumberAvailable == 0 {
+		availableConditionReason = "NoAPIServerNode"
```
deads2k: NoAPIServerPod

pkg/operator/workload_controller_openshiftapiserver_v311_00.go
deads2k: > check all these first because the last set wins and the daemonset being missing is more important to report.

nm, I see

deads2k: If you don't have *all* the APIservices (this can happen on errors), you must indicate that availability is unknown.

deads2k: > If you don't have _all_ the APIservices (this can happen on errors), you must indicate that availability is unknown.

add a test for this.  It's a correctness problem.

pkg/operator/workload_controller_openshiftapiserver_v311_00.go
```diff
@@ -244,8 +277,8 @@ func manageOpenShiftAPIServerDaemonSet_v311_00_to_latest(client appsclientv1.Dae
 	return resourceapply.ApplyDaemonSet(client, recorder, required, resourcemerge.ExpectedDaemonSetGeneration(required, generationStatus), forceRollingUpdate)
 }
 
-func manageAPIServices_v311_00_to_latest(client apiregistrationv1client.APIServicesGetter) error {
-	apiServices := []schema.GroupVersion{
+func manageAPIServices_v311_00_to_latest(client apiregistrationv1client.APIServicesGetter) ([]*apiregistrationv1.APIService, error) {
+	apiServiceGroups := []schema.GroupVersion{
```
deads2k: `apiServiceGroupVersions` if you're going to rename

pkg/operator/workload_controller_openshiftapiserver_v311_00.go
deads2k: This is still outstanding.  You would reporting failing and you *also* need to report unavailable.

pkg/operator/workload_controller_openshiftapiserver_v311_00.go
sanchezl: The case where `actualDaemonSet == nil` is handle above.

pkg/operator/workload_controller_openshiftapiserver_v311_00.go
sanchezl: > If you don't have _all_ the APIservices (this can happen on errors), you must indicate that availability is unknown.

* It appears that I either get all `APIServices` , or none if there is **any** error invoking `manageAPIServices_v311_00_to_latest`. 
* When there is an error invoking `manageAPIServices_v311_00_to_latest`, the error will appear in the `WorkloadFailng` condition's message and the `Available` condition would of been set to `Unknown` in the preceding `switch` statement.

pkg/operator/workload_controller_openshiftapiserver_v311_00.go
```diff
@@ -89,49 +89,95 @@ func syncOpenShiftAPIServer_v311_00_to_latest(c OpenShiftAPIServerOperator, orig
...
 	}
 
 	// manage status
-	if actualDaemonSet.Status.NumberAvailable > 0 {
+	var availableConditionReason string
+	var availableConditionMessages []string
+	availableConditionStatus := operatorv1.ConditionFalse
+
+	switch {
+	case actualDaemonSet == nil:
```
deads2k: remove this case

pkg/operator/workload_controller_openshiftapiserver_v311_00.go
```diff
@@ -89,49 +89,95 @@ func syncOpenShiftAPIServer_v311_00_to_latest(c OpenShiftAPIServerOperator, orig
...
+		availableConditionStatus = operatorv1.ConditionUnknown
+		availableConditionReason = "DaemonAvailabilityUnknown"
+		availableConditionMessages = append(availableConditionMessages, "daemonset/apiserver.openshift-apiserver: status unknown")
+	case actualDaemonSet.Status.NumberAvailable > 0 && len(actualAPIServices) == 0:
+		availableConditionStatus = operatorv1.ConditionUnknown
+		availableConditionReason = "APIServiceAvailabilityUnknown"
+		availableConditionMessages = append(availableConditionMessages, "apiservice/*: status unknown")
+	}
+
+	if actualDaemonSet != nil && actualDaemonSet.Status.NumberAvailable == 0 {
```
deads2k: if you're using a switch/case, put this there.

pkg/operator/workload_controller_openshiftapiserver_v311_00.go
```diff
@@ -89,49 +89,95 @@ func syncOpenShiftAPIServer_v311_00_to_latest(c OpenShiftAPIServerOperator, orig
...
+	var availableConditionReason string
+	var availableConditionMessages []string
+	availableConditionStatus := operatorv1.ConditionFalse
+
+	switch {
+	case actualDaemonSet == nil:
+		availableConditionStatus = operatorv1.ConditionUnknown
+		availableConditionReason = "DaemonAvailabilityUnknown"
+		availableConditionMessages = append(availableConditionMessages, "daemonset/apiserver.openshift-apiserver: status unknown")
+	case actualDaemonSet.Status.NumberAvailable > 0 && len(actualAPIServices) == 0:
```
deads2k: make these two different dimensions, same as you did for the two progresing conditions

pkg/operator/workload_controller_openshiftapiserver_v311_00.go
```diff
@@ -89,49 +89,95 @@ func syncOpenShiftAPIServer_v311_00_to_latest(c OpenShiftAPIServerOperator, orig
...
+				break
+			}
+		}
+	}
+	switch {
+	case len(availableConditionMessages) == 1:
 		v1helpers.SetOperatorCondition(&operatorConfig.Status.Conditions, operatorv1.OperatorCondition{
-			Type:   operatorv1.OperatorStatusTypeAvailable,
-			Status: operatorv1.ConditionTrue,
+			Type:    operatorv1.OperatorStatusTypeAvailable,
```
deads2k: yeah, use false down here.

pkg/operator/workload_controller_openshiftapiserver_v311_00.go
```diff
@@ -89,49 +89,95 @@ func syncOpenShiftAPIServer_v311_00_to_latest(c OpenShiftAPIServerOperator, orig
...
 	// If the daemonset is up to date and the operatorConfig are up to date, then we are no longer progressing
-	var progressingMessages []string
-	if actualDaemonSet.ObjectMeta.Generation != actualDaemonSet.Status.ObservedGeneration {
-		progressingMessages = append(progressingMessages, fmt.Sprintf("daemonset/apiserver.openshift-operator: observed generation is %d, desired generation is %d.", actualDaemonSet.Status.ObservedGeneration, actualDaemonSet.ObjectMeta.Generation))
-	}
-	if operatorConfig.ObjectMeta.Generation != operatorConfig.Status.ObservedGeneration {
-		progressingMessages = append(progressingMessages, fmt.Sprintf("openshiftapiserveroperatorconfigs/instance: observed generation is %d, desired generation is %d.", operatorConfig.Status.ObservedGeneration, operatorConfig.ObjectMeta.Generation))
-	}
-
-	if len(progressingMessages) == 0 {
```
deads2k: why did you touch this

pkg/operator/workload_controller_openshiftapiserver_v311_00.go
```diff
@@ -89,49 +89,95 @@ func syncOpenShiftAPIServer_v311_00_to_latest(c OpenShiftAPIServerOperator, orig
...
+	if actualDaemonSet == nil || operatorConfig == nil {
 		v1helpers.SetOperatorCondition(&operatorConfig.Status.Conditions, operatorv1.OperatorCondition{
 			Type:   operatorv1.OperatorStatusTypeProgressing,
-			Status: operatorv1.ConditionFalse,
+			Status: operatorv1.ConditionUnknown,
```
deads2k: this looks wrong.

pkg/operator/workload_controller_openshiftapiserver_v311_00.go
```diff
@@ -89,49 +89,95 @@ func syncOpenShiftAPIServer_v311_00_to_latest(c OpenShiftAPIServerOperator, orig
...
 		})
 	} else {
-		v1helpers.SetOperatorCondition(&operatorConfig.Status.Conditions, operatorv1.OperatorCondition{
-			Type:    operatorv1.OperatorStatusTypeProgressing,
-			Status:  operatorv1.ConditionTrue,
-			Reason:  "DesiredStateNotYetAchieved",
-			Message: strings.Join(progressingMessages, "\n"),
-		})
+		var progressingMessages []string
+		if actualDaemonSet.ObjectMeta.Generation != actualDaemonSet.Status.ObservedGeneration {
```
deads2k: put these back.

pkg/operator/workload_controller_openshiftapiserver_v311_00.go
```diff
@@ -89,49 +89,95 @@ func syncOpenShiftAPIServer_v311_00_to_latest(c OpenShiftAPIServerOperator, orig
...
+	var availableConditionMessages []string
+	availableConditionStatus := operatorv1.ConditionFalse
+
+	switch {
+	case actualDaemonSet == nil:
+		availableConditionStatus = operatorv1.ConditionUnknown
+		availableConditionReason = "DaemonAvailabilityUnknown"
+		availableConditionMessages = append(availableConditionMessages, "daemonset/apiserver.openshift-apiserver: status unknown")
+	case actualDaemonSet.Status.NumberAvailable > 0 && len(actualAPIServices) == 0:
+		availableConditionStatus = operatorv1.ConditionUnknown
```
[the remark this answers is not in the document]
sanchezl: @deads2k just confirming that this, alone with getting rid of the 'DaemonAvailabilityUnknown` case above, leaves only  `True` & `False` as statuses for the 'Available' condition?

pkg/operator/workload_controller_openshiftapiserver_v311_00.go
sanchezl: to avoid panic when `actualDaemonSet == nil`

pkg/operator/workload_controller_openshiftapiserver_v311_00.go
sanchezl: As with `Available`, I figured if we could not get the `actualDaemonSet` then we can't tell is we are progressing or not.

pkg/operator/workload_controller_openshiftapiserver_v311_00.go
sanchezl: The test `actualDaemonSet.Status.NumberAvailable > 0` lets us know that we actually attempted to manage the `apiServices`

PR #89, opened 2018-12-10
copy cluster image CAs to api server configmap

bparees:
1) watch the image.config.openshift.io and the configmap it references as a source of additional CAs to trust for image import.

2) sync that configmap into the openshift-apiserver namespace and mount the configmap into the apiserver pod.  CAs will be consumed by the imageimport logic inside the apiserver (work to be done in origin).

### DIFF
--- a/bindata/v3.11.0/openshift-apiserver/ds.yaml
+++ b/bindata/v3.11.0/openshift-apiserver/ds.yaml
@@ -41,6 +41,8 @@ spec:
           name: client-ca
         - mountPath: /var/run/configmaps/etcd-serving-ca
           name: etcd-serving-ca
+        - mountPath: /var/run/configmaps/image-import-ca
+          name: image-import-ca
         - mountPath: /var/run/secrets/etcd-client
           name: etcd-client
         - mountPath: /var/run/secrets/serving-cert
@@ -67,6 +69,10 @@ spec:
       - name: etcd-serving-ca
         configMap:
           name: etcd-serving-ca
+      - name: image-import-ca
+        configMap:
+          name: image-import-ca
+          optional: true
       - name: etcd-client
         secret:
           secretName: etcd-client
--- a/pkg/operator/workload_controller_openshiftapiserver_v311_00.go
+++ b/pkg/operator/workload_controller_openshiftapiserver_v311_00.go
@@ -6,6 +6,7 @@ import (
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/sets"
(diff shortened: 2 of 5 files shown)


bindata/v3.11.0/openshift-apiserver/ds.yaml
```diff
@@ -41,6 +41,8 @@ spec:
           name: client-ca
         - mountPath: /var/run/configmaps/etcd-serving-ca
           name: etcd-serving-ca
+        - mountPath: /var/run/configmaps/image-import-ca
+          name: image-import-ca
```
mfojtik: should this be allowMissing ?

pkg/operator/workload_controller_openshiftapiserver_v311_00.go
```diff
@@ -69,8 +71,14 @@ func syncOpenShiftAPIServer_v311_00_to_latest(c OpenShiftAPIServerOperator, orig
...
 
+	imageImportCAModified := false
+	imageImportCAModified, err = manageOpenShiftAPIServerImageImportCA_v311_00_to_latest(c.openshiftConfigClient, c.kubeClient.CoreV1(), c.eventRecorder)
+	if err != nil {
+		errors = append(errors, fmt.Errorf("%q: %v", "client-ca", err))
+	}
+
 	forceRollingUpdate = forceRollingUpdate || operatorConfig.ObjectMeta.Generation != operatorConfig.Status.ObservedGeneration
-	forceRollingUpdate = forceRollingUpdate || configMapModified || etcdModified || clientCAModified
+	forceRollingUpdate = forceRollingUpdate || configMapModified || etcdModified || clientCAModified || imageImportCAModified
```
mfojtik: @deads2k this is getting uglier

pkg/operator/workload_controller_openshiftapiserver_v311_00.go
```diff
@@ -69,8 +71,14 @@ func syncOpenShiftAPIServer_v311_00_to_latest(c OpenShiftAPIServerOperator, orig
 		errors = append(errors, fmt.Errorf("%q: %v", "client-ca", err))
 	}
 
+	imageImportCAModified := false
```
mfojtik: why not just `imageImportCAModified, err :=` ?

pkg/operator/workload_controller_openshiftapiserver_v311_00.go
```diff
@@ -172,6 +180,32 @@ func manageOpenShiftAPIServerClientCA_v311_00_to_latest(client coreclientv1.Core
 	return caChanged, nil
 }
 
+func manageOpenShiftAPIServerImageImportCA_v311_00_to_latest(openshiftConfigClient openshiftconfigclientv1.ConfigV1Interface, client coreclientv1.CoreV1Interface, recorder events.Recorder) (bool, error) {
+	const imageImportCA = "image-import-ca"
```
mfojtik: move this to global and rename to `imageImportCAName`

pkg/operator/workload_controller_openshiftapiserver_v311_00.go
bparees: leftover from how i iterated my way there. will fix.

pkg/operator/workload_controller_openshiftapiserver_v311_00.go
bparees: I will do this, but please note that i followed the pattern from 10 lines up:

```
        const apiserverClientCA = "client-ca"
 	_, caChanged, err := resourceapply.SyncConfigMap(client, recorder, kubeAPIServerNamespaceName, apiserverClientCA, targetNamespaceName, apiserverClientCA)
```

bindata/v3.11.0/openshift-apiserver/ds.yaml
bparees: ?

I don't see that as a field on volumeMounts.  I don't think pods care if the configmap you're mounting doesn't exist, isn't that a key thing?

bparees:
@mfojtik thanks for the review, comments addressed.

bindata/v3.11.0/openshift-apiserver/ds.yaml
bparees: ah, was needed on the volume section, not volumeMount.  added.

bparees:
@mfojtik bump

PR #91, opened 2018-12-11
Revert "add clusteroperator resource to CVO payload"

abhinavdahiya:
This reverts commit a07dc5caaa4bdb13513a28bb2382464520580ab1.

For more info https://github.com/openshift/cluster-kube-apiserver-operator/pull/167

/cc @deads2k


abhinavdahiya:
> /lgtm
> 
> Toaybe unstick things, but I'd really like to see a precise bad status so we know what we have to fix

https://github.com/openshift/release/pull/2376 this should help get visibility into the clusteroperator objects on ci runs.

PR #97, opened 2018-12-18
add service monitor resources for operator

deads2k:
/assign @mfojtik


manifests/0000_90_cluster-openshift-apiserver-operator_03_servicemonitor.yaml
```diff
@@ -0,0 +1,27 @@
...
+    metricRelabelings:
+    - action: drop
+      regex: etcd_(debugging|disk|request|server).*
+      sourceLabels:
+      - __name__
+    port: https
+    scheme: https
+    tlsConfig:
+      # TODO where do you mount the service-ca.crt?
+      insecureSkipVerify: true
```
mfojtik: probably need a follow up issue?

mfojtik:
/lgtm

we probably need issue so we won't forget to make the collection secure

PR #103, opened 2019-01-09
 add rollout when aggregation and client CAs change

deads2k:
The openshift-apiserver needs to get the updated trust bundles when we rotate them.  This is a very simple way to do it.  We can do it more effectively and reliably with content copy after we fix some configuration.

/assign @sttts @mfojtik


pkg/operator/workload_controller_openshiftapiserver_v311_00.go
```diff
@@ -265,6 +265,11 @@ func manageOpenShiftAPIServerConfigMap_v311_00_to_latest(kubeClient kubernetes.I
 			Namespace: targetNamespaceName,
 			Name:      "serving-cert",
 		},
+		resourcehash.ObjectReference{
```
deads2k: this is the meat here.  It calculates a hash based on content and changes to that hash will cause a config change, which will cause a rollout

deads2k:
controller manager again.  We need https://github.com/openshift/cluster-openshift-controller-manager-operator/pull/50

PR #105, opened 2019-01-10
force clear the kube finalizer when kube resources are gone

deads2k:
this allows us to delete the openshift-apiserver namespace.

When running an aggregated API on the platform, you delete the namespace hosting the aggregated API.  Doing that the namespace controller starts by doing complete discovery and then deleting all objects, but pods have a grace period, so it deletes the rest and requeues.  The ns controller starts again and does a complete discovery and.... fails.  The failure means it refuses to complete the cleanup.  Now, we don't *actually* want to delete the resoruces from our aggregated API, only the server plus config
if we remove the apiservices to unstick it, GC will start cleaning everything.  For now, we can unbork 4.0, but clearing the finalizer after the pod and daemonset we created are gone.

/assign @mfojtik @sttts

### DIFF
--- a/pkg/operator/finalizer_controller.go
+++ b/pkg/operator/finalizer_controller.go
@@ -1,0 +1,174 @@
+package operator
+
+import (
+	"fmt"
+	"reflect"
+	"time"
+
+	"github.com/golang/glog"
+
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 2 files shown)


deads2k:
This allows us to actually delete the `openshift-apiserver` namespace to "reset" without getting wedged.

sttts:
Is this some kind of hack? Removing a foreign finalizer....

deads2k:
explained. code was ok'd in slack.

PR #107, opened 2019-01-12
wait for kubeapiserver to be ready

deads2k:
If we wait for the kube-apiserver to be available via cluster operator status, then we are assured that it has written out the authentication ca bundle configmap.  We can't really do anything without a ready kube-apiserver, so this should be zero impact.

@smarterclayton @sttts I really should have thought of this first.

/assign @sttts


deads2k:
Green, going for it

PR #118, opened 2019-01-21
project: fix project config observation

mfojtik:
We were setting wrong field in openshiftcontrolplane.OpenShiftAPIServerConfig which caused the apiserver pods to go into crashloop as we were not able to decode the output... 

This change should set the right field (also add handlers for project).


pkg/operator/configobservation/project/observe_projects.go
```diff
@@ -13,8 +13,12 @@ import (
 )
 
 var (
-	projectRequestMessagePath      = []string{"projectConfig", "projectRequestMessage"}
-	projectRequestTemplateNamePath = []string{"projectConfig", "projectRequestTemplate", "name"}
+	// This represents a JSON path for openshiftcontrolplane/v1.OpenShiftAPIServerConfig
```
sttts: nit: s/for/inside/

sttts:
description: + "and we didn't watch projects".

/lgtm
/approve

smarterclayton:
Can we root cause how this merged with passing tests?  Was it just that we got really unlucky and it passed randomly?

mfojtik:
@smarterclayton 

1) no test that exercised this on live cluster
2) the CR (projects.config.openshift.io/v1) was never created (it will using cluster-config-operator)

I tried to test this manually today and found the unstructured path was wrong because it nuked my openshift apiserver.

PR #120, opened 2019-01-21
ingress: add event to ingress observer when domain change

mfojtik:
@ironcladlou this add event so when bad things start to happen in your observer we can see them...

### DIFF
--- a/pkg/operator/configobservation/ingresses/observe_ingresses.go
+++ b/pkg/operator/configobservation/ingresses/observe_ingresses.go
@@ -1,6 +1,8 @@
 package ingresses
 
 import (
+	"reflect"
+
 	"github.com/golang/glog"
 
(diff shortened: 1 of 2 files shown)


pkg/operator/configobservation/ingresses/observe_ingresses.go
```diff
@@ -51,5 +53,11 @@ func ObserveIngressDomain(genericListers configobserver.Listers, recorder events
 		}
 	}
 
+	if reflect.DeepEqual(routingDomain, currentRoutingDomain) {
```
Miciah: `routingDomain` and `currentRoutingDomain` are just strings; why use `reflect.DeepEqual` rather than `==`?

mfojtik: just to make sure when they change we are safe

PR #121, opened 2019-01-21
move openshift apiserver later in CVO payload

deads2k:
nothing relies on openshift until late in the startup


sttts:
Nobody reviewed this?

PR #123, opened 2019-01-22
fix cache mutation issue and version reporting

deads2k:
this fixes a cache mutation problem we had checking syncs.  I think it will help with stability.


sttts:
Please link the library-go PR this is based on.

PR #125, opened 2019-01-22
 add a check that actually attempts to hit the openshift api discovery

deads2k:
Long ago, past me hit a problem in the aggregator that I suspected was openapi taking a long time.  We would transition an apiservice to available, but the proxy wasn't ready.  This does an actual check against the kube-apiserver to wait for them all to respond.

ref https://github.com/openshift/origin/blob/release-3.11/pkg/oc/clusteradd/componentinstall/readiness_apigroup.go

### DIFF
--- a/pkg/operator/starter.go
+++ b/pkg/operator/starter.go
@@ -5,6 +5,8 @@ import (
 	"os"
 	"time"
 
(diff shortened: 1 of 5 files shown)


deads2k:
this has a good shot at fixing 3374 and 3373

PR #129, opened 2019-01-22
always requeue on operator status that needs to be rechecked

deads2k:
Because the apiservices are listed as available, but aren't actually available, we need to requeue when our status isn't good.  This isn't quite like a resync. We only do this until we settle.

@eparis


eparis:
I know I'm not sufficiently intelligent to review this one   :)

eparis:
ok, actually, maybe I can figure it out.
/lgtm

deads2k:
just prometheus

PR #130, opened 2019-01-23
Configure auditing

sttts:
... to be on by default and log to /var/log/openshift-apiserver on the host, in analogy to kube-apiserver.

Prerequisite: https://github.com/openshift/origin/pull/21831


sttts:
/hold

until https://github.com/openshift/origin/pull/21831 merged.

deads2k:
/lgtm

@sttts I'm sorry this needs a rebase.  Long term I'd like to see us become compatible with kube flags, but given an operator we can switch at our leisure.

PR #132, opened 2019-01-24
Switch to openshift/api API

deads2k:
This switches us to the openshift/api API

/assign @damemi


damemi:
/hold
merges should wait until rebase lands

deads2k:
/retest

statefulset, but it installed

PR #134, opened 2019-01-26
correct available condition when no missing API services

sanchezl:
Fix a code path where we end up with an incorrect condition of Available==false.


deads2k:
/hold

I'm not seeing it. We return an empty list which we append, thus appending nothing.  What's the bug?

sanchezl:
Closing as it should not be needed.

PR #137, opened 2019-02-06
update to manage versions correctly

deads2k:
This updates our operator to correctly report operand and operator versions in clusteroperator.status. It does this by relying the substitution value (@smarterclayton you're going to catch me on that right?) and (in the case of library-go), the watchable, thread-safe version application logic.

It also relies of the clever use of CVO merge logic (I think) to avoid overwriting all configmap keys so that I effectively keep a history of all the levels I've installed in the past.

@smarterclayton @bparees @ironcladlou 
@openshift/sig-master

### DIFF
--- a/pkg/operator/workloadcontroller/workload_controller.go
+++ b/pkg/operator/workloadcontroller/workload_controller.go
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/openshift/library-go/pkg/operator/status"
+
 	"github.com/openshift/cluster-openshift-apiserver-operator/pkg/operator/operatorclient"
 
 	"github.com/golang/glog"
@@ -39,6 +41,7 @@ const (
 
 type OpenShiftAPIServerOperator struct {
 	targetImagePullSpec string
+	versionRecorder     status.VersionGetter
 
 	operatorConfigClient    operatorv1client.OpenShiftAPIServersGetter
 	openshiftConfigClient   openshiftconfigclientv1.ConfigV1Interface
@@ -54,6 +57,7 @@ type OpenShiftAPIServerOperator struct {
 
 func NewWorkloadController(
 	targetImagePullSpec string,
+	versionRecorder status.VersionGetter,
 	operatorConfigInformer operatorv1informers.OpenShiftAPIServerInformer,
 	kubeInformersForOpenShiftAPIServerNamespace kubeinformers.SharedInformerFactory,
 	kubeInformersForEtcdNamespace kubeinformers.SharedInformerFactory,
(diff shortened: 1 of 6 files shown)


deads2k:
> So why make every operator responsible for tracking it itself?

It was a compromise between @smarterclayton and myself in the design to be able to accommodate level driven version checking.  I think most/all components need that level check though.

pkg/operator/workloadcontroller/workload_controller.go
```diff
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/openshift/library-go/pkg/operator/status"
+
 	"github.com/openshift/cluster-openshift-apiserver-operator/pkg/operator/operatorclient"
```
soltysh: You like messing up our imports :stuck_out_tongue:

smarterclayton:
The CVO can't ensure that the component X was able to capture all inputs Y
for versions {N}, nor can it ensure that component X has taken all required
actions until component X tells the CVO via the version field.

Think of the version field as a nonce to your deployment - you don't have
to care what it is, but you are required to write it to CO status when you
are sure that everything that needs to be done has been done at least
once.  I'm not aware of any other way to perform that signaling without
designing a new API object type per version that the CVO creates that the
CO watches to write to, which is effectively no different than CO status
anyway.

bparees:
I was only suggesting that the CVO track the history of versions not that the CVO be responsible for indicating when a particular version is successfully rolled out.

deads2k:
rebased, retagging

deads2k:
openshift api blipped.  I saw that in one pull earlier this week trying again

deads2k:
```
object.ObjectMeta.ResourceVersion:
  a: "34598"
  b: "34766"
object.Status.Phase:
  a: "Failed"
  b: "Succeeded"
object.Status.Conditions[0].Reason:
  a: ""
  b: "PodCompleted"
object.Status.Conditions[1].Reason:
  a: "ContainersNotReady"
  b: "PodCompleted"
object.Status.Conditions[1].Message:
  a: "containers with unready status: [deployment]"
  b: ""
object.Status.Conditions[2].Reason:
  a: "ContainersNotReady"
  b: "PodCompleted"
object.Status.Conditions[2].Message:
  a: "containers with unready status: [deployment]"
  b: ""
Expected
    <bool>: true
to be false
			
```


weird dc failure in https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_cluster-openshift-apiserver-operator/137/pull-ci-openshift-cluster-openshift-apiserver-operator-master-e2e-aws/681

deads2k:
/retest

Known networking flakes

deads2k:
Known flake

PR #142, opened 2019-02-12
Switch operator to use protobuf client for non-CR resources #250

mfojtik:
This builds on [openshift/library-go#233](https://github.com/openshift/library-go/pull/233) and switches the controller loops that deal with non-CR resources to client-go with protobuf enforced.

/cc @deads2k
/cc @smarterclayton


pkg/operator/starter.go
```diff
@@ -76,7 +76,8 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
 	resourceSyncController, err := resourcesynccontroller.NewResourceSyncController(
 		operatorClient,
 		kubeInformersForNamespaces,
-		kubeClient,
+		v1helpers.CachedConfigMapGetter(kubeClient.CoreV1(), kubeInformersForNamespaces),
```
mfojtik: @deads2k ptal

PR #150, opened 2019-02-20
Set openshift-apiserver privileged for audit to be able to write to /var/log/openshift-apiserver

soltysh:
/assign @deads2k


soltysh:
Added comment per @mfojtik request, self-tagging back.

mfojtik:
known flakes, merging by hand.

mfojtik:
heh, no merge button :/

PR #156, opened 2019-02-26
default the internal registry hostname to what we expect it to be, save a restart

bparees:


pkg/operator/configobservation/images/observe_images.go
```diff
@@ -39,6 +39,11 @@ func ObserveInternalRegistryHostname(genericListers configobserver.Listers, reco
 	configImage, err := listers.ImageConfigLister.Get("cluster")
 	if errors.IsNotFound(err) {
 		glog.Warningf("image.config.openshift.io/cluster: not found")
+		// if we have no cluster image config object, default the internal registry hostname to what we expect it will be
```
deads2k: do you want this same logic down on an else on line 54?

bparees:
after discussion with @deads2k we agree that defaulting this to anything is the wrong thing to do as it implies a registry is available when one may not be.

PR #163, opened 2019-03-05
use dynamic reloading of certs to avoid rolling out new deployment versions

deads2k:
Since https://github.com/openshift/origin/pull/22229 merged, we don't have to reload to get certs.

/hold

holding until I complete testing locally.


deads2k:
worked like a champ

PR #167, opened 2019-03-06
cmd: fix operator component name

mfojtik:
/cc @deads2k


deads2k:
I bet something explodes

/lgtm

PR #168, opened 2019-03-13
workload: set conditions when kube apiserver is missing

mfojtik:
Set the condition properly when the pre-requirements are missing. This allow us to debug a case when the workload controller wedge and never create the namespace.

The second commit will add event handlers to kube apiserver informer for faster requeue when the kube apiserver config is available.

/cc @sttts 
/cc @deads2k

### DIFF
--- a/pkg/operator/workloadcontroller/workload_controller.go
+++ b/pkg/operator/workloadcontroller/workload_controller.go
@@ -4,10 +4,6 @@ import (
 	"fmt"
 	"time"
 
-	"github.com/openshift/library-go/pkg/operator/status"
-
-	"github.com/openshift/cluster-openshift-apiserver-operator/pkg/operator/operatorclient"
-
 	"github.com/golang/glog"
 
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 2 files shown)


mfojtik:
@smarterclayton this should help debug why the workload controller is wedged and not creating the namespace for the operand.

pkg/operator/workloadcontroller/workload_controller.go
```diff
@@ -103,43 +112,38 @@ func (c OpenShiftAPIServerOperator) sync() error {
...
-		}
-		return nil
-	default:
-		c.eventRecorder.Warningf("ManagementStateUnknown", "Unrecognized operator management state %q", operatorConfig.Spec.ManagementState)
+	if !management.IsOperatorManaged(operatorConfig.Spec.ManagementState) {
 		return nil
 	}
 
+	cond := operatorv1.OperatorCondition{
+		Type:   "PreRequirementNotReady",
```
derekwaynecarr: nit: prerequisite is probably better than prerequirement

mfojtik: sure, fixed.

PR #170, opened 2019-03-22
Increase minimal-shutdown-duration to 35s to cope with slowly converging SDN

sttts:
We still see that endpoints disappear right away on deletion, but requests hit the API server.


sttts:
We use the endpoints directly in the aggregator, without going through networking. I think we can safely keep the short termination period.

PR #172, opened 2019-03-26
Debug handler

damemi:
Depends on https://github.com/openshift/library-go/pull/321

Adds call to handler function to serve resource controller sync rules

### DIFF
--- a/pkg/operator/resourcesynccontroller/resourcesynccontroller.go
+++ b/pkg/operator/resourcesynccontroller/resourcesynccontroller.go
@@ -1,6 +1,8 @@
 package resourcesynccontroller
 
 import (
+	"net/http"
+
 	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
 
 	"github.com/openshift/library-go/pkg/operator/events"
@@ -15,7 +17,7 @@ func NewResourceSyncController(
 	kubeInformersForNamespaces v1helpers.KubeInformersForNamespaces,
 	configMapsGetter corev1client.ConfigMapsGetter,
 	secretsGetter corev1client.SecretsGetter,
-	eventRecorder events.Recorder) (*resourcesynccontroller.ResourceSyncController, error) {
+	eventRecorder events.Recorder) (*resourcesynccontroller.ResourceSyncController, http.Handler, error) {
 
 	resourceSyncController := resourcesynccontroller.NewResourceSyncController(
 		operatorConfigClient,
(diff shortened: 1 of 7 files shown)


pkg/operator/starter.go
```diff
@@ -148,6 +148,10 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
 
 	configUpgradeableController := unsupportedconfigoverridescontroller.NewUnsupportedConfigOverridesController(operatorClient, ctx.EventRecorder)
 
+	if ctx.Server != nil {
+		ctx.Server.Handler.NonGoRestfulMux.HandleFunc("/syncRules", resourceSyncController.ServeSyncRules)
```
deads2k: This looks off.  

`/debug/controllers/resourcesync`

pkg/operator/starter.go
```diff
@@ -148,6 +148,10 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
 
 	configUpgradeableController := unsupportedconfigoverridescontroller.NewUnsupportedConfigOverridesController(operatorClient, ctx.EventRecorder)
 
+	if ctx.Server != nil {
+		ctx.Server.Handler.NonGoRestfulMux.Handle("/debug/controllers/resourcesync", debugHandler)
```
deads2k: I suspect that we'll want `handleprefix` in the end, but ok.

damemi:
ping @deads2k is this good to merge for now?

pkg/operator/resourcesynccontroller/resourcesynccontroller.go
```diff
@@ -28,26 +30,26 @@ func NewResourceSyncController(
...
 	if err := resourceSyncController.SyncConfigMap(
 		resourcesynccontroller.ResourceLocation{Namespace: operatorclient.TargetNamespace, Name: "aggregator-client-ca"},
 		resourcesynccontroller.ResourceLocation{Namespace: operatorclient.GlobalMachineSpecifiedConfigNamespace, Name: "kube-apiserver-aggregator-client-ca"},
 	); err != nil {
-		return nil, err
+		return nil, nil, err
 	}
 
-	return resourceSyncController, nil
+	return resourceSyncController, resourcesynccontroller.NewDebugHandler(resourceSyncController), nil
```
sttts: why do we need a return value for the handler if it is only built from the already returned `resourceSyncController` and `NewDebugHandler`?

damemi:
@sttts based on feedback from David we decided to explicitly return the
debug handler (see
https://github.com/openshift/library-go/pull/321#issuecomment-476762957).
Am I misinterpreting something?

damemi:
@sttts the other option is importing both `resourcesynccontroller` packages from this operator and from library-go into `starter.go`, which seems unnecessary since it would only be imported for this function

pkg/operator/starter.go
sttts: Can ctx.Server be nil?

PR #176, opened 2019-04-08
add test-e2e make target

sanchezl:


Makefile
```diff
@@ -33,6 +33,12 @@ clean:
 	$(RM) ./cluster-openshift-apiserver-operator
 .PHONY: clean
 
+GO_TEST_PACKAGES :=./pkg/... ./cmd/...
+
+.PHONY: test-e2e
+test-e2e: GO_TEST_PACKAGES :=./test/e2e/...
```
deads2k: @tnozicka is this normally "WHAT"?

deads2k:
just a quesiton

/lgtm
/hold

holding for @tnozicka's answer

test/library/client.go
```diff
@@ -0,0 +1,19 @@
+package library
+
+import (
+	"fmt"
```
tnozicka: nit: separate stdlib import with `\n`

Makefile
tnozicka: Yes, in kube shell magic. There are different default for different targets so I've went for discrete and descriptive names. We would likely needed those anyways, but I am open to wrapping WHAT around them if you'd like, I get that it would add simplicity when observed externally. But that's a change for library-go, this is just setting the correct value for a variable library-go expects in target `test-unit`.
https://github.com/openshift/library-go/blob/6401c32c6160d8a689c1033b8be294b3df4f1e0a/alpha-build-machinery/make/lib/golang.mk#L6-L16
https://github.com/openshift/library-go/blob/44a56c866078ec596987e464de21cd078bc06a28/alpha-build-machinery/make/targets/golang/test-unit.mk#L5-L4

/hold cancel

PR #192, opened 2019-04-23
permanently swap to openshift-etcd

deads2k:
followup to https://github.com/openshift/machine-config-operator/pull/648 that makes the operator only respect new, not old

/assign @mfojtik 
@openshift/sig-master


bindata/v3.11.0/openshift-apiserver/defaultconfig.yaml
```diff
@@ -43,6 +43,9 @@ auditConfig:
...
         omitStages:
           - "RequestReceived"
+storageConfig:
+  urls:
+    - https://etcd.openshift-etcd.svc:2379
 apiServerArguments:
   minimal-shutdown-duration:
-  - 3s # give SDN some time to converge
\ No newline at end of file
+  - 3s # give SDN some time to converge
```
deads2k: @mfojtik I'm thinking we should bump this up to the 35 we found for kube-apiserver

sttts: openshift-apiserver works through endpoint IP directly, not via iptables or ELBs. So if our informers in kube-aggregator work, we should be safe here with 3s.

sttts:
As follow-up of https://github.com/openshift/machine-config-operator/pull/648

/lgtm

PR #198, opened 2019-04-30
use standard operator helper to update status

deads2k:
found debugging SCC.  This slows rollout

### DIFF
--- a/pkg/operator/workloadcontroller/workload_controller.go
+++ b/pkg/operator/workloadcontroller/workload_controller.go
@@ -4,10 +4,6 @@ import (
 	"fmt"
 	"time"
 
-	"github.com/openshift/library-go/pkg/operator/status"
-
-	"github.com/openshift/cluster-openshift-apiserver-operator/pkg/operator/operatorclient"
-
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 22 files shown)


pkg/operator/workloadcontroller/workload_controller.go
```diff
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
```
mfojtik: nit: move down

mfojtik:
/lgtm
/hold

one nit, feel free to self-tag

pkg/operator/workloadcontroller/workload_controller_openshiftapiserver.go
```diff
@@ -216,22 +215,31 @@ func syncOpenShiftAPIServer_v311_00_to_latest(c OpenShiftAPIServerOperator, orig
 		c.versionRecorder.SetVersion("openshift-apiserver", operandVersion)
 
 	}
-	if !equality.Semantic.DeepEqual(operatorConfig.Status, originalOperatorConfig.Status) {
-		if _, err := c.operatorConfigClient.OpenShiftAPIServers().UpdateStatus(operatorConfig); err != nil {
-			return false, err
-		}
+
+	currOperatorStatus, _, err := v1helpers.UpdateStatus(
```
jboyd01: if UpdateStatus() determines new status == old status, won't currOperatorStatus will be returned as nil? If yes, we need to add a check for nil and an early return otherwise expect issues at line 236

PR #201, opened 2019-05-03
use standard object update

deads2k:
fix up from https://github.com/openshift/cluster-openshift-apiserver-operator/pull/198 .  already lgtmd once

### DIFF
--- a/pkg/operator/starter.go
+++ b/pkg/operator/starter.go
@@ -93,6 +93,7 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
 		kubeInformersForNamespaces.InformersFor(operatorclient.GlobalUserSpecifiedConfigNamespace),
 		apiregistrationInformers,
 		configInformers,
+		operatorClient,
 		operatorConfigClient.OperatorV1(),
 		configClient.ConfigV1(),
 		kubeClient,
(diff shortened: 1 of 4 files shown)


sanchezl:
The `status.dockerImageRepository` field seems to be flapping between `""` and the expected value in language build imagestreams (e.g. `ruby`, `nodejs`, `pearl`, etc...). I say flapping because we see it get the needed value in `ruby` (for example) than fail getting it from `nodejs`, when on the next iteration of the `WaitForOpenShiftNamespaceImageStreams` loop, `ruby` no longer contains the needed value.

awgreene:
Hello - as the build cop today this PR seemed to run into an issue with `ImageStreams` roughly 8 times in the past 24 hours: https://search.svc.ci.openshift.org/?search=Failed+to+import+expected+imagestreams&maxAge=168h&context=2&type=all

There might be an issue that won't be resolved by rerunning the test.

vrutkovs:
/hold

Retests of this PR have been failing for a while - see https://openshift-gce-devel.appspot.com/builds/origin-ci-test/pr-logs/pull/openshift_cluster-openshift-apiserver-operator/201/pull-ci-openshift-cluster-openshift-apiserver-operator-master-e2e-aws/

The test failure is related to an image-registry/samples update few weeks ago. I think a rebase would fix that

enj:
/uncc

@stlaz @sttts @mfojtik

PR #210, opened 2019-07-10
Use new crd-schema-gen Makefile pattern

damemi:
This follows the same approach as 
* https://github.com/openshift/cluster-kube-apiserver-operator/pull/514
* https://github.com/openshift/cluster-openshift-controller-manager-operator/pull/105


Dockerfile
```diff
@@ -1,7 +1,7 @@
-FROM registry.svc.ci.openshift.org/openshift/release:golang-1.11 AS builder
+FROM registry.svc.ci.openshift.org/openshift/release:golang-1.12 AS builder
 WORKDIR /go/src/github.com/openshift/cluster-openshift-apiserver-operator
 COPY . .
-RUN go build ./cmd/cluster-openshift-apiserver-operator
+RUN GODEBUG=tls13=1 go build ./cmd/cluster-openshift-apiserver-operator
```
deads2k: what does this do?

damemi: @deads2k [this change](https://github.com/openshift/library-go/blob/master/pkg/crypto/crypto.go#L34-L39) in library-go adds something that uses TLS 1.3. According to the comment in that block it's enabled by this build flag in go 1.12

damemi: Go docs say it will be enabled by default in go 1.13

PR #216, opened 2019-07-25
Start observing the Proxy object

stlaz:
Adds an observer for the Proxy object, pulls its fields and passes
them to the environment of the deployed openshift-apiserver container.


stlaz:
Pulled in latest lib-go for the Proxy observer

pkg/operator/workloadcontroller/workload_controller_openshiftapiserver_v311_00.go
```diff
@@ -21,6 +22,7 @@ import (
 	apiregistrationv1 "k8s.io/kube-aggregator/pkg/apis/apiregistration/v1"
 	apiregistrationv1client "k8s.io/kube-aggregator/pkg/client/clientset_generated/clientset/typed/apiregistration/v1"
 
+	"github.com/ghodss/yaml"
```
sttts: nit: order

pkg/operator/workloadcontroller/workload_controller_openshiftapiserver_v311_00.go
```diff
@@ -397,3 +409,17 @@ func resourceSelectorForCLI(obj runtime.Object) string {
 	}
 	return kind + group + "/" + name
 }
+
+func proxyMapToEnvVars(proxyConfig map[string]string) []corev1.EnvVar {
+	envVars := []corev1.EnvVar{}
```
sttts: move down where it is used, and return nil in the if-statement. Easier to read...

pkg/operator/workloadcontroller/workload_controller_openshiftapiserver_v311_00.go
```diff
@@ -326,6 +328,16 @@ func manageOpenShiftAPIServerDaemonSet_v311_00_to_latest(client appsclientv1.Dae
...
 
+	var observedConfig map[string]interface{}
+	if err := yaml.Unmarshal(operatorConfig.Spec.ObservedConfig.Raw, &observedConfig); err != nil {
+		return nil, false, fmt.Errorf("failed to unmarshal the observedConfig: %v", err)
+	}
+	proxyConfig, _, err := unstructured.NestedStringMap(observedConfig, "workloadcontroller", "proxy")
+	if err != nil {
+		return nil, false, fmt.Errorf("couldn't get the proxy config from observedConfig: %v", err)
+	}
+	required.Spec.Template.Spec.Containers[0].Env = append(required.Spec.Template.Spec.Containers[0].Env, proxyMapToEnvVars(proxyConfig)...)
```
sttts: should be append these to all containers? The cert syncher does not need it. But this kind of kind has bitten us before. I would prefer it to be generic.

stlaz: agree, that's probably the better thing to do

stlaz:
addressed the comments

stlaz:
/retest
servicemonitor race

stlaz:
/retest
The servicemonitor race seems to appear quite often

stlaz:
Hm, there's actually a bug in this PR, ProxyLister won't be callable.

PR #221, opened 2019-08-14
Bug 1737683: Sort the EnvVar from Proxy string map

stlaz:


pkg/operator/workloadcontroller/workload_controller_openshiftapiserver_v311_00.go
```diff
@@ -434,5 +434,6 @@ func proxyMapToEnvVars(proxyConfig map[string]string) []corev1.EnvVar {
 		envVars = append(envVars, corev1.EnvVar{Name: k, Value: v})
 	}
 
+	sort.Slice(envVars, func(i, j int) bool { return envVars[i].Name < envVars[j].Name })
```
sttts: add a comment why we sort? I guess to make it stable and avoid updates?

stlaz:
we probably don't need to do this for the DS, closing

PR #222, opened 2019-08-14
fix make images

sallyom:
This PR fixes `make images`, it requires `image name` now. 

(`make images` is only used in local development, AFAIK)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,13 +9,15 @@ include $(addprefix ./vendor/github.com/openshift/library-go/alpha-build-machine
 	targets/openshift/images.mk \
 )
 
+IMAGE_REGISTRY :=registry.svc.ci.openshift.org
+
 # This will call a macro called "build-image" which will generate image specific targets based on the parameters:
 # $0 - macro name
-# $1 - target suffix
-# $2 - Dockerfile path
-# $3 - context directory for image build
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -11,11 +11,12 @@ include $(addprefix ./vendor/github.com/openshift/library-go/alpha-build-machine
...
-# $1 - target suffix
-# $2 - Dockerfile path
-# $3 - context directory for image build
+# $1 - target name
+# $2 - image ref
+# $3 - Dockerfile path
+# $4 - context directory for image build
 # It will generate target "image-$(1)" for builing the image an binding it as a prerequisite to target "images".
-$(call build-image,origin-$(GO_PACKAGE),./Dockerfile,.)
+$(call build-image,origin-$(go_build_binaries),docker.io/origin-$(go_build_binaries),./Dockerfile,.)
```
tnozicka: here is a template for the new one https://github.com/openshift/oc/blob/55a3aad4d40cba59d5367f15b850c36bafba4ed1/Makefile#L45

sallyom: noted here: https://github.com/openshift/cluster-kube-apiserver-operator/pull/551#discussion_r317760633   the `make images` cmd is not useful as/is for developers, and ART does not use the make cmd AFIAK for building images.  hm... why do we include it?

PR #223, opened 2019-08-14
use non-static operator client

sanchezl:
Fixes a small typo where a `StaticPodOperatorClient` is created instead of an `OperatorClient`.

### DIFF
--- a/pkg/operator/starter.go
+++ b/pkg/operator/starter.go
@@ -60,7 +60,7 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
 	apiregistrationInformers := apiregistrationinformers.NewSharedInformerFactory(apiregistrationv1Client, 10*time.Minute)
 	configInformers := configinformers.NewSharedInformerFactory(configClient, 10*time.Minute)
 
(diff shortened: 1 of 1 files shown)


pkg/operator/starter.go
```diff
@@ -60,7 +60,7 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
 	apiregistrationInformers := apiregistrationinformers.NewSharedInformerFactory(apiregistrationv1Client, 10*time.Minute)
 	configInformers := configinformers.NewSharedInformerFactory(configClient, 10*time.Minute)
 
-	operatorClient, dynamicInformers, err := genericoperatorclient.NewStaticPodOperatorClient(ctx.KubeConfig, operatorv1.GroupVersion.WithResource("openshiftapiservers"))
+	operatorClient, dynamicInformers, err := genericoperatorclient.NewClusterScopedOperatorClient(ctx.KubeConfig, operatorv1.GroupVersion.WithResource("openshiftapiservers"))
```
sttts: is this wrong in all of our operators?

soltysh: From my understanding `StaticPodOperator` is meant only for operators managing static pods. Since osa-o is not this change is legit.

PR #226, opened 2019-08-20
Inject the trusted-ca-bundle into openshift-apiserver pods

stlaz:
cc @sttts @mfojtik


manifests/0000_30_openshift-apiserver-operator_07_deployment.yaml
```diff
@@ -43,6 +43,8 @@ spec:
         env:
         - name: IMAGE
           value: quay.io/openshift/origin-openshift-apiserver:v4.2
+        - name: OPERATOR_IMAGE
+          value: docker.io/openshift/origin-cluster-openshift-apiserver-operator:v4.0
```
mfojtik: this does not seem right?

pkg/operator/workloadcontroller/workload_controller_openshiftapiserver_v311_00.go
```diff
@@ -277,6 +283,33 @@ func manageOpenShiftAPIServerImageImportCA_v311_00_to_latest(openshiftConfigClie
 	}
 }
 
+func ensureOpenShiftAPIServerTrustedCA_v311_00_to_latest(client coreclientv1.CoreV1Interface, recorder events.Recorder) error {
+	const injectCAAnnotationKey = "config.openshift.io/inject-trusted-cabundle"
```
mfojtik: no more consts! :-)

pkg/operator/workloadcontroller/workload_controller_openshiftapiserver_v311_00.go
```diff
@@ -277,6 +283,33 @@ func manageOpenShiftAPIServerImageImportCA_v311_00_to_latest(openshiftConfigClie
...
+	cm, err := cmCLient.Get("trusted-ca-bundle", metav1.GetOptions{})
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			_, err = cmCLient.Create(required)
+		}
+		return err
+	}
+
+	// check that the inject label exists and is set to "true"
+	injectCA, ok := cm.Labels[injectCAAnnotationKey]
```
mfojtik: convert this to `func hasInjectCAAnnotation(cm) bool`

bindata/v3.11.0/openshift-apiserver/ds.yaml
```diff
@@ -101,6 +134,12 @@ spec:
       - name: serving-cert
         secret:
           secretName: serving-cert
+      - name: trusted-ca-bundle
```
mfojtik: should this be optional?

manifests/0000_30_openshift-apiserver-operator_07_deployment.yaml
stlaz: that's taken from image-references, I can fix it in both places but not only one of them. What would the correct value be? `s/docker/quay/`?

bindata/v3.11.0/openshift-apiserver/ds.yaml
```diff
@@ -77,6 +80,36 @@ spec:
             scheme: HTTPS
             port: 8443
             path: healthz
+      - name: openshift-apiserver-watchdog
+        command: ["cluster-openshift-apiserver-operator", "file-watcher-watchdog"]
+        args:
+          - --namespace=$(POD_NAMESPACE)
+          - --process-name=openshift-apiserver
+          - --termination-grace-period=30s
```
mfojtik: 60s

bindata/v3.11.0/openshift-apiserver/ds.yaml
stlaz: Yes, given a message from proxy people yesterday, non-optional CMs like this may prolong upgrades since the network operator may start injecting the CMs too late

pkg/operator/workloadcontroller/workload_controller_openshiftapiserver_v311_00.go
stlaz: Consts are helpful if you have clumsy fingers like I do :) But alright, this one will appear only in one spot in this fucntion anyway

stlaz:
/test e2e-aws
/hold
I wonder whether the errors we're seeing could be caused by mounting an empty configmap to trusted store dir

bindata/v3.11.0/openshift-apiserver/ds.yaml
mfojtik: Right answer was no :-) We want to require this.

mfojtik:
@deads2k @sttts just so we are clear... this add mechanism that will force all three openshift apiservers to restart at the same time without rolling new revision out (daemonset). I'm fine with it, because this just substitute the dynamic cert reloading. From what I tested manually, the restart is pretty fast, but aggregator might notice it.

stlaz:
The tests passing now only proves that the CM injection is kind of flawed, will add a two step process to make sure the CM always has data not to wedge the cluster by deleting its system trust store when mounting a CM with no data...

stlaz:
/hold cancel
there's a mechanism that ensures the mounted CM should have content in it already

danehans:
> there's a mechanism that ensures the mounted CM should have content in it already

@stlaz operators should tolerate the trusted ca configmap being empty. There may be cases where your operator starts up before the network operator that's responsible for fulfilling the configmap injection request.

stlaz:
@danehans this is not about the operator, but the operand, which can never have empty trust store as that is a potential outage. It would also make all e2e tests very flaky.

stlaz:
What I could do is wait for a different mount location to be filled with the file that’d be copied to the location we need in a bash wrapper, yet I don’t think I could so that for KAS because I don’t think anything would fill the hostPath mount with the data from the config map

stlaz:
`ReadDaemonSetV1OrDie` can't read `-|` for long strings, `- |` is fine though, pushed.

stlaz:
looks like the CI got stuck?
/refresh

bindata/v3.11.0/openshift-apiserver/ds.yaml
stlaz: Apparently yes was still the good answer :)) There's people who want us to have it optional for the reasons above, this is now handled by bash guard in the pod's `commands`

danehans: optional is needed to support 4.1>4.2 upgrades.

manifests/0000_30_openshift-apiserver-operator_07_deployment.yaml
sttts: fixed

stlaz:
/hold
Latest logs show that openshift-apiserver most probably broke, you can also see that the watchdog can’t find the PID

stlaz:
From openshift-apiserver logs: “Waiting for watchdog........”

bindata/v3.11.0/openshift-apiserver/ds.yaml
```diff
@@ -36,9 +37,21 @@ spec:
         terminationMessagePolicy: FallbackToLogsOnError
         image: ${IMAGE}
         imagePullPolicy: IfNotPresent
-        command: ["openshift-apiserver", "start"]
+        command: ["/bin/bash", "-ec"]
         args:
-        - "--config=/var/run/configmaps/config/config.yaml"
+          - |
+            echo $$ > /var/run/watchdog/pid
```
stlaz: You probably noticed, these dollar signs seem to be escaped, so they won’t be interpreted.

I wonder if we could either use realpath of /proc/self or get the value of either /proc/self/stat{,us}. Starts to be a bit hackish.

sttts: they should be escaped and passed to bash, which then interprets them.

sttts: yaml doesn't escape I hope

sttts: but I agree that `W0827 11:29:32.124287      29 cmd.go:211] Unable to parse pid file /var/run/watchdog/pid: strconv.Atoi: parsing "$": invalid syntax` is convincing.

stlaz: I’m just assuming given the watchdog reports “/var/run/watchdog/pid: strconv.Atoi: parsing "$": invalid syntax”

sttts: `$$$$` does the trick. Is that yaml escaping of dollar?

stlaz: Hard to tell, I did not think dollar sign has a special meaning for yaml files :/

sttts: try `kubectl run --restart=Never -i -t busybox --image busybox -- /bin/sh -ec 'echo $$$$'`

sttts:
Upgrade flakes seem unrelated.

/retest
/hold cancel

mfojtik:
/hold

@deads2k had opinions about using the watchdog, I think we should just rollout new deployment when proxy change.

mfojtik:
/hold cancel

Alright, we will go with this to unblock 4.2, but there should be BZ created with follow up to revert this approach and just use the existing operator mechanism to rollout new version.

sttts:
Follow-up bugzilla https://bugzilla.redhat.com/show_bug.cgi?id=1746375

PR #230, opened 2019-08-30
Bug 1747260: manifests: move trusted-ca into right spot

mfojtik:
This moves the configmap into right spot (operator namespace). Operator will then copy it.

NOTE: Operator will copy it with the inject label, which means both this operator and networking operator will manage the config map. Since it should always have the right content, I don't see this as a big problem, but it breaks the pattern of "only one controller manage".

### DIFF
--- a/pkg/operator/workloadcontroller/workload_controller_openshiftapiserver_v311_00.go
+++ b/pkg/operator/workloadcontroller/workload_controller_openshiftapiserver_v311_00.go
@@ -81,11 +81,6 @@ func syncOpenShiftAPIServer_v311_00_to_latest(c OpenShiftAPIServerOperator, orig
 		reasonsForForcedRollingUpdate = append(reasonsForForcedRollingUpdate, "modified: "+resourceSelectorForCLI(imageImportCAModifiedObject))
 	}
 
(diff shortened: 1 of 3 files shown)


mfojtik:
@deads2k ptal

sttts:
/hold

I don't like this. Why is it not a big problem to have multiple operators modifying the same object? This asks for races.

pkg/operator/workloadcontroller/workload_controller_openshiftapiserver_v311_00.go
```diff
@@ -282,28 +277,6 @@ func manageOpenShiftAPIServerImageImportCA_v311_00_to_latest(openshiftConfigClie
 	}
 }
 
-func ensureOpenShiftAPIServerTrustedCA_v311_00_to_latest(client coreclientv1.CoreV1Interface, recorder events.Recorder) error {
```
sttts: why is this removed?

sttts:
Before merging anything else, this must be smoke tested in a real cluster.

pkg/operator/workloadcontroller/workload_controller_openshiftapiserver_v311_00.go
mfojtik: not needed, the resource syncer take care of this

stlaz:
/close
in favour of #231 and #232

PR #231, opened 2019-09-02
Bug 1747260: Fix trusted-ca-bundle CM syncing

stlaz:
The trusted-ca-bundle was never created in the operator NS where
we were trying to sync it from. Have it be created by the CVO in
the operator NS and sync only its data (therefore not the injection
label) to the operand's NS.

### DIFF
--- a/pkg/operator/workloadcontroller/workload_controller_openshiftapiserver_v311_00.go
+++ b/pkg/operator/workloadcontroller/workload_controller_openshiftapiserver_v311_00.go
@@ -2,6 +2,7 @@ package workloadcontroller
 
 import (
 	"fmt"
+	"reflect"
 	"sort"
 	"strings"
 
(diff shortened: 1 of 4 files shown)


stlaz:
/retest
eh, tests passed but appear red

pkg/operator/workloadcontroller/workload_controller_openshiftapiserver_v311_00.go
```diff
@@ -282,26 +283,42 @@ func manageOpenShiftAPIServerImageImportCA_v311_00_to_latest(openshiftConfigClie
...
-		_, err = cmCLient.Update(cm)
+	// it's optional, don't sync it if it has no data -> prevent trust store being removed upon startup/upgrade
+	if len(sourceCM.Data) == 0 {
+		return nil
+	}
+
+	targetClient := client.ConfigMaps(operatorclient.TargetNamespace)
+	imageCM, err := targetClient.Get("trusted-ca-bundle", metav1.GetOptions{})
+	if apierrors.IsNotFound(err) {
+		_, err = targetClient.Create(&corev1.ConfigMap{
```
sttts: is targetClient cached?

PR #233, opened 2019-09-25
Update deps to 1.16

marun:


marun:
Failing e2e jobs appear to be consistent, suggesting that this PR is the cause.

sttts:
Seeing `remote error: tls: bad certificate` in the openshift-apiserver logs.

marun:
Tests are passing, but PR needs to be held until https://github.com/openshift/library-go/pull/538 merges and the change is vendored in.

marun:
@sttts Ready

pkg/operator/workloadcontroller/workload_controller_openshiftapiserver_v311_00.go
```diff
@@ -463,6 +463,7 @@ func manageAPIServices_v311_00_to_latest(client apiregistrationv1client.APIServi
 				VersionPriority:      15,
 			},
 		}
+		apiregistrationv1.SetDefaults_ServiceReference(obj.Spec.Service)
```
sttts: oh, you are lucky here. Defaults are considered private to the API server. But we make them accessible by accident.

PR #235, opened 2019-10-18
add correct label to ds config initContainer

haircommander:
CRI-O 1.16 made a change to how selinux labels were handled that increased security. In doing so, pod specs need to enable selinuxRelabel.

In the case of this initContainer, it is trying to edit the permissions of a file on the host system, which would typically be considered security flaw, and thus selinux is correctly denying it.

Add the type spc_t, which allows the container to modify the host file

Signed-off-by: Peter Hunt <pehunt@redhat.com>

### DIFF
--- a/bindata/v3.11.0/openshift-apiserver/ds.yaml
+++ b/bindata/v3.11.0/openshift-apiserver/ds.yaml
@@ -28,6 +28,9 @@ spec:
           image: ${IMAGE}
           imagePullPolicy: IfNotPresent
           command: ['sh', '-c', 'chmod 0700 /var/log/openshift-apiserver']
+          securityContext:
+            selinuxOptions:
+              type: spc_t
           volumeMounts:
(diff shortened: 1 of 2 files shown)


haircommander:
note: this PR is largely to use cluster-bot to test whether this is the needed change

haircommander:
cc @umohnani8 @mrunalp

bindata/v3.11.0/openshift-apiserver/ds.yaml
```diff
@@ -28,6 +28,9 @@ spec:
           image: ${IMAGE}
           imagePullPolicy: IfNotPresent
           command: ['sh', '-c', 'chmod 0700 /var/log/openshift-apiserver']
+          securityContext:
+            selinuxOptions:
```
mrunalp: should be `privileged: true`

haircommander:
Closing in favor of #236

PR #236, opened 2019-10-18
add correct label to ds config initContainer

umohnani8:
CRI-O 1.16 made a change to how selinux labels were handled that increased security. In doing so, pod specs need to enable selinuxRelabel. In the case of this initContainer, it is trying to edit the permissions of a file on the host system, which is a security flaw, and thus selinux is denying it.

Add the type spc_t, which allows the container to modify the host file

Signed-off-by: Peter Hunt <pehunt@redhat.com>


tnozicka:
/approve
to unblock cri-o 1.16 but in the long term I don't think the container should be privileged to access mounted path, the mounting of the hostPath seems to be a place to enforce permissions/labels

PR #237, opened 2019-10-23
Add encryption support

sttts:


pkg/operator/resourcesynccontroller/resourcesynccontroller.go
```diff
@@ -50,6 +50,12 @@ func NewResourceSyncController(
 	); err != nil {
 		return nil, nil, err
 	}
+	if err := resourceSyncController.SyncSecret(
+		resourcesynccontroller.ResourceLocation{Namespace: operatorclient.TargetNamespace, Name: "encryption-config"},
```
p0lyn0mial: won't it be synced in `NewRevisionLabelPodDeployer?

pkg/operator/starter.go
```diff
@@ -135,6 +160,43 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
...
+	nodeProvider := DaemonSetNodeProvider{
+		TargetNamespaceDaemonSetInformer: kubeInformersForNamespaces.InformersFor(operatorclient.TargetNamespace).Apps().V1().DaemonSets(),
+		NodeInformer:                     kubeInformersForNamespaces.InformersFor("").Core().V1().Nodes(),
+	}
+	deployer, err := encryptiondeployer.NewRevisionLabelPodDeployer("revision", operatorclient.TargetNamespace, kubeInformersForNamespaces, resourceSyncController, kubeClient.CoreV1(), kubeClient.CoreV1(), nodeProvider)
+	if err != nil {
+		return err
+	}
+
+	encryptionControllers, err := encryption.NewControllers(
```
p0lyn0mial: should we change `NewControllers` to accept an alternative implementation of the `prune` ctrl?

pkg/operator/workloadcontroller/workload_controller_openshiftapiserver_v311_00.go
```diff
@@ -365,35 +367,22 @@ func manageOpenShiftAPIServerDaemonSet_v311_00_to_latest(
...
-		case "${OPERATOR_IMAGE}":
-			required.Spec.Template.Spec.InitContainers[i].Image = operatorImagePullSpec
-		default:
-			if strings.Contains(img, "$") {
-				return nil, false, fmt.Errorf("invalid image reference %q", img)
-			}
-		}
+	tmpl := v311_00_assets.MustAsset("v3.11.0/openshift-apiserver/ds.yaml")
+
+	r := strings.NewReplacer(
```
p0lyn0mial: that must be the bit that will replace `encryption-config-${REVISION}`

pkg/operator/starter.go
```diff
@@ -76,6 +87,20 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
 		return err
 	}
 
+	revisionController := revisioncontroller.NewRevisionController(
```
p0lyn0mial: ok, this controller will create secrets with revisions

pkg/operator/prunecontroller/prune_controller.go
```diff
@@ -0,0 +1,232 @@
...
+	eventRecorder events.Recorder
+}
+
+const (
+	pruneControllerWorkQueueKey = "key"
+	numOldRevisionsToPreserve   = 5
+)
+
+// NewPruneController creates a new pruning controller
+func NewPruneController(
```
p0lyn0mial: why do we need a new controller?

pkg/operator/latestrevisionclient.go
```diff
@@ -0,0 +1,65 @@
...
+// OpenshiftDeploymentLatestRevisionClient is an LatestRevisionClient implementation for openshift-apiserver..
+type OpenshiftDeploymentLatestRevisionClient struct {
+	v1helpers.OperatorClient
+	TypedClient operatorv1client.OpenShiftAPIServersGetter
+}
+
+var _ revisioncontroller.LatestRevisionClient = OpenshiftDeploymentLatestRevisionClient{}
+
+func (c OpenshiftDeploymentLatestRevisionClient) GetLatestRevisionState() (*operatorv1.OperatorSpec, *operatorv1.OperatorStatus, int32, string, error) {
+	o, err := c.TypedClient.OpenShiftAPIServers().Get("cluster", metav1.GetOptions{})
```
p0lyn0mial: ah, so we will take the revision from the operator.

pkg/operator/resourcesynccontroller/resourcesynccontroller.go
sttts: yep. Will remove thi sone.

pkg/operator/starter.go
sttts: Not sure we should couple the prune controller with the encryption controllers without a need.

pkg/operator/workloadcontroller/workload_controller_openshiftapiserver_v311_00.go
sttts: yes

pkg/operator/prunecontroller/prune_controller.go
sttts: because we have none? The one from static pods is super complex and does much much more. I looked into generalizing it, but decided it is not worth it.

pkg/operator/latestrevisionclient.go
sttts: Yes, the same way the static pod variant works, just different operator config types. And as we know the types, I just use the typed client here instead of dynamic client magic.

sttts:
/hold

until smoke test worked.

pkg/operator/prunecontroller/prune_controller.go
mfojtik: @sttts will this controller move to library-go ?

PR #238, opened 2019-10-25
encryption: remove key secret finalizer before pruning

sttts:
The key secret have a finalizer. This must be removed before deletion.

### DIFF
--- a/pkg/operator/prunecontroller/prune_controller.go
+++ b/pkg/operator/prunecontroller/prune_controller.go
@@ -8,15 +8,19 @@ import (
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/wait"
 	corev1informer "k8s.io/client-go/informers/core/v1"
 	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/tools/cache"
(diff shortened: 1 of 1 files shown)


sttts:
/hold 

until manually tested (we don't have e2e tests for this).

pkg/operator/prunecontroller/prune_controller.go
```diff
@@ -96,7 +100,32 @@ func (c *PruneController) sync() error {
...
+				}
+				newFinalizers = append(newFinalizers, f)
+			}
+			s.Finalizers = newFinalizers
+
+			_, err = c.secretGetter.Secrets(s.Namespace).Update(s)
+			return err
+		})
+
+		if err := c.secretGetter.Secrets(s.Namespace).Delete(s.Name, nil); err != nil && !errors.IsNotFound(err) {
```
p0lyn0mial: ok, so regardless of the status of the previous operation (success, failure) remove the secret, why not :)

pkg/operator/prunecontroller/prune_controller.go
```diff
@@ -96,7 +100,32 @@ func (c *PruneController) sync() error {
...
+			newFinalizers := make([]string, 0, len(s.Finalizers))
+			for _, f := range s.Finalizers {
+				if f == encryptionsecrets.EncryptionSecretFinalizer {
+					continue
+				}
+				newFinalizers = append(newFinalizers, f)
+			}
+			s.Finalizers = newFinalizers
+
+			_, err = c.secretGetter.Secrets(s.Namespace).Update(s)
```
p0lyn0mial: if `len(newFinalizers) == len(s.Finalizers)` do not update, wdyt ?

sttts: done

pkg/operator/prunecontroller/prune_controller.go
sttts: Worth a try.

PR #248, opened 2019-11-05
Bug 1767156: Fix "Multipleavailable" reason in the code 

sohankunkerkar:
I have come up with this change after having some discussion around the Bugzilla with @stlaz.

### DIFF
--- a/pkg/operator/workloadcontroller/workload_controller_openshiftapiserver_v311_00.go
+++ b/pkg/operator/workloadcontroller/workload_controller_openshiftapiserver_v311_00.go
@@ -119,25 +119,25 @@ func syncOpenShiftAPIServer_v311_00_to_latest(c OpenShiftAPIServerOperator, orig
 	}
 
 	// manage status
-	var availableConditionReason string
+	var availableConditionReasons []string
 	var availableConditionMessages []string
 
 	switch {
 	case actualDaemonSet == nil:
-		availableConditionReason = "NoDaemon"
+		availableConditionReasons = append(availableConditionReasons, "NoDaemon")
 		availableConditionMessages = append(availableConditionMessages, "daemonset/apiserver.openshift-apiserver: could not be retrieved")
 	case actualDaemonSet.Status.NumberAvailable == 0:
-		availableConditionReason = "NoAPIServerPod"
+		availableConditionReasons = append(availableConditionReasons, "NoAPIServerPod")
 		availableConditionMessages = append(availableConditionMessages, "no openshift-apiserver daemon pods available on any node.")
 	case actualDaemonSet.Status.NumberAvailable > 0 && len(actualAPIServices) == 0:
-		availableConditionReason = "NoRegisteredAPIServices"
+		availableConditionReasons = append(availableConditionReasons, "NoRegisteredAPIServices")
 		availableConditionMessages = append(availableConditionMessages, "registered apiservices could not be retrieved")
 	}
 	for _, apiService := range actualAPIServices {
 		for _, condition := range apiService.Status.Conditions {
 			if condition.Type == apiregistrationv1.Available {
 				if condition.Status == apiregistrationv1.ConditionFalse {
-					availableConditionReason = "APIServiceNotAvailable"
+					availableConditionReasons = append(availableConditionReasons, "APIServiceNotAvailable")
 					availableConditionMessages = append(availableConditionMessages, fmt.Sprintf("apiservice/%v: not available: %v", apiService.Name, condition.Message))
 				}
 				break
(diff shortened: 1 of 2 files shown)


pkg/operator/workloadcontroller/workload_controller_openshiftapiserver_v311_00_test.go
```diff
@@ -393,8 +393,17 @@ func TestAvailableStatus(t *testing.T) {
...
+				actualReasons := strings.Split(condition.Reason, "\n")
+				a := make([]string, len(tc.expectedReasons))
+				b := make([]string, len(actualReasons))
+				copy(a, tc.expectedReasons)
+				copy(b, actualReasons)
+				sort.Strings(a)
+				sort.Strings(b)
+				if !equality.Semantic.DeepEqual(a, b) {
+					t.Error("\n" + diff.ObjectDiff(a, b))
+				}
```
stlaz: Why not to compare this as strings instead? You know what those should look like

pkg/operator/workloadcontroller/workload_controller_openshiftapiserver_v311_00_test.go
```diff
@@ -280,9 +280,9 @@ func TestAvailableStatus(t *testing.T) {
 			},
 		},
 		{
-			name:           "MultipleAPIServiceNotAvailable",
-			expectedStatus: operatorv1.ConditionFalse,
-			expectedReason: "Multiple",
+			name:            "MultipleAPIServiceNotAvailable",
+			expectedStatus:  operatorv1.ConditionFalse,
+			expectedReasons: []string{"APIServiceNotAvailable1", "APIServiceNotAvailable2"},
```
stlaz: These don't get indexed in the code so this will never pass. Omit the numbers.

pkg/operator/workloadcontroller/workload_controller_openshiftapiserver_v311_00_test.go
sohankunkerkar: I think I understood what's missing here. I should have used sort function only `expectedReasons` array because the `actualReasons` array would already be in sorted order.

pkg/operator/workloadcontroller/workload_controller_openshiftapiserver_v311_00_test.go
sohankunkerkar: Fixed.

sohankunkerkar:
> The changes lgtm, but please squash your commits to a single one.

Done!. Thanks for reviewing.

stlaz:
/retitle Bug 1767156: Fix "Multipleavailable" reason in the code

pkg/operator/workloadcontroller/workload_controller_openshiftapiserver_v311_00_test.go
```diff
@@ -393,8 +393,11 @@ func TestAvailableStatus(t *testing.T) {
 			if condition.Status != tc.expectedStatus {
 				t.Error(diff.ObjectGoPrintSideBySide(condition.Status, tc.expectedStatus))
 			}
-			if tc.expectedReason != "" && condition.Reason != tc.expectedReason {
-				t.Error(diff.ObjectGoPrintSideBySide(condition.Reason, tc.expectedReason))
+			expectedReasons := strings.Join(tc.expectedReasons, "\n")
+			if len(expectedReasons) > 0 && condition.Reason != expectedReasons {
+				{
```
stlaz: no need to nest this block any further I think

sohankunkerkar: Fixed.

pkg/operator/workloadcontroller/workload_controller_openshiftapiserver_v311_00.go
```diff
@@ -119,25 +119,25 @@ func syncOpenShiftAPIServer_v311_00_to_latest(c OpenShiftAPIServerOperator, orig
...
 
 	// manage status
-	var availableConditionReason string
+	var availableConditionReasons []string
 	var availableConditionMessages []string
 
 	switch {
 	case actualDaemonSet == nil:
-		availableConditionReason = "NoDaemon"
+		availableConditionReasons = append(availableConditionReasons, "NoDaemon")
```
deads2k: messy. We should have a condition for each one.  I think this good enough to backport, but in 4.3, they should be individual and we should clear the existing "Available".

pkg/operator/workloadcontroller/workload_controller_openshiftapiserver_v311_00.go
```diff
@@ -151,19 +151,21 @@ func syncOpenShiftAPIServer_v311_00_to_latest(c OpenShiftAPIServerOperator, orig
...
 
+	sort.Sort(sort.StringSlice(availableConditionReasons))
+
 	switch {
 	case len(availableConditionMessages) == 1:
 		v1helpers.SetOperatorCondition(&operatorConfig.Status.Conditions, operatorv1.OperatorCondition{
 			Type:    operatorv1.OperatorStatusTypeAvailable,
 			Status:  operatorv1.ConditionFalse,
-			Reason:  availableConditionReason,
+			Reason:  availableConditionReasons[0],
```
sttts: this panics

PR #250, opened 2019-11-06
auditConfig: Log write operations of Identities with RequestResponse

jhrozek:
In order to be able to match a user from IDP with a username which is
normally logged in the audit logs, we need to have the creation of the
Identity API resource logged.

This patch changes the hardcoded auditConfig to log "witethe "identities"
resource of the "user.openshift.io" group at the RequestResponse level.

### DIFF
--- a/bindata/v3.11.0/openshift-apiserver/defaultconfig.yaml
+++ b/bindata/v3.11.0/openshift-apiserver/defaultconfig.yaml
@@ -42,6 +42,13 @@ auditConfig:
           - "/api*" # Wildcard matching.
           - "/version"
           - "/healthz"
+      # Log the full Identity API resource object so that the audit trail
+      # allows us to match the username with the IDP identity.
+      - level: RequestResponse
+        verbs: ["create", "update", "patch"]
+        resources:
+          - group: "user.openshift.io"
+            resources: ["identities"]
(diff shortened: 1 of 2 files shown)


jhrozek:
CC @stlaz @sttts - we talked about this on Slack last week.

deads2k:
I'm fine with this change.

bindata/v3.11.0/openshift-apiserver/defaultconfig.yaml
```diff
@@ -42,6 +42,11 @@ auditConfig:
           - "/api*" # Wildcard matching.
           - "/version"
           - "/healthz"
+      - level: RequestResponse
```
sttts: We have an explanation for the reasoning behind the rules everywhere else. Let's add one here too.

jhrozek: Sure, thanks for the review. Are you OK with the comment I added in the latest iteration?

sttts: looks good

sttts: looks good

PR #252, opened 2019-11-08
Add default ClusterRole and ClusterRoleBinding for motd

JAORMX:
With https://github.com/openshift/oc/pull/67 came the addition of the
capability to display a message of the day via the `oc` command. This
message of the day serves the purpose of presenting a legal notice
that's required for regulatory compliance.

In order for this legal notice to be usable out of the box, we need to
grant folks read-only access to it. This PR adds that by creating a
default ClusterRole and ClusterRoleBinding that enables this.

### DIFF
--- a/manifests/0000_30_openshift-apiserver-operator_04_motd_role.yaml
+++ b/manifests/0000_30_openshift-apiserver-operator_04_motd_role.yaml
@@ -1,0 +1,22 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: motd-reader
+rules:
+- apiGroups: [""]
+  resources: ["configmaps"]
+  resourceNames: ["motd"]
+  verbs: ["get"]
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: read-motd-global
+subjects:
+- kind: Group
+  name: system:authenticated
+  apiGroup: rbac.authorization.k8s.io
+roleRef:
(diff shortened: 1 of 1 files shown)


sttts:
@deads2k @stlaz @marun @mfojtik @soltysh where do we want to put default policy rules in the future, in this case to make oc motd work?

soltysh:
> @deads2k @stlaz @marun @mfojtik @soltysh where do we want to put default policy rules in the future, in this case to make oc motd work?

Why do we need a separate role, isn't the default read-only `openshift` namespace sufficient?

JAORMX:
@soltysh did something change recently though? cause I was sure we added the motd to the "openshift" namespace because we had that read-only role. But I don't seem to find the role, and it certainly doesn't work as it used to. Our QE stumbled upon that, needing to add the ClusterRole to get the motd to work,

JAORMX:
Yep, just tried it again and the clusterRole is needed, unfortunately. Perhaps this is a regression?

soltysh:
> Yep, just tried it again and the clusterRole is needed, unfortunately. Perhaps this is a regression?

Can we update that other role to include configmaps? `openshift` namespace is meant as a read-only resource by default

JAORMX:
@soltysh sure! I just don't know what that other role is. I actually have been looking at where the `openshift` namespaces and associated roles are created. But I haven't found that yet :confused: 

@sttts @deads2k @soltysh any idea where that would be?

JAORMX:
Closing this in favor of https://github.com/openshift/openshift-apiserver/pull/42

PR #261, opened 2019-11-21
load test for migration

p0lyn0mial:


pkg/operator/starter.go
```diff
@@ -44,7 +44,10 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
 	if err != nil {
 		return err
 	}
-	dynamicClient, err := dynamic.NewForConfig(ctx.KubeConfig)
+	dynamicClientConfig := dynamic.ConfigFor(ctx.KubeConfig)
+	dynamicClientConfig.Burst = 40
```
p0lyn0mial: we need to experiment and find some good numbers for QPS/Burst

p0lyn0mial:
this pull requires https://github.com/openshift/library-go/pull/595

pkg/operator/starter.go
sttts: What does kube-storage-version-migrator use?

sttts: Please rename `dynamicClient` to `migrationClient` to make clear it is not meant for anything else than migration.

test/e2e-encryption/encryption_test.go
```diff
@@ -13,6 +13,7 @@ import (
 )
 
 func TestEncryptionTypeIdentity(t *testing.T) {
+	t.SkipNow()
```
sttts: WIP?

pkg/operator/starter.go
p0lyn0mial: > What does kube-storage-version-migrator use?

it seems like they use default settings https://github.com/kubernetes-sigs/kube-storage-version-migrator/blob/master/cmd/migrator/app/migrator.go#L58

PR #275, opened 2019-12-09
create apiservice controller

deads2k:
Separate the apiservices into a different control loop so we can re-use it for the oauth-apiserver

/assign @p0lyn0mial @sanchezl @sohankunkerkar


pkg/operator/apiservicecontroller/apiservice_controller.go
```diff
@@ -0,0 +1,232 @@
...
+		return errors.NewAggregate(errs)
+	}
+	if len(availableConditionMessages) > 0 {
+		sort.Sort(sort.StringSlice(availableConditionMessages))
+		return fmt.Errorf(strings.Join(availableConditionMessages, "\n"))
+	}
+
+	// if the apiservices themselves check out ok, try to actually hit the discovery endpoints.  We have a history in clusterup
+	// of something delaying them.  This isn't perfect because of round-robining, but let's see if we get an improvement
+	if len(availableConditionMessages) == 0 && c.kubeClient.Discovery().RESTClient() != nil {
```
stlaz: `availableConditionMessages` will always be of len 0 here

pkg/operator/apiservicecontroller/apigroup.go
```diff
@@ -0,0 +1,78 @@
...
+		curr := coordinate{namespace: apiService.Spec.Service.Namespace, name: apiService.Spec.Service.Name}
+		exists := false
+		for _, j := range coordinates {
+			if j == curr {
+				exists = true
+				break
+			}
+		}
+		if !exists {
+			coordinates = append(coordinates)
```
stlaz: `append(coordinates, curr)`?

PR #276, opened 2019-12-09
separate finalizer cleanup controller to allow re-use

deads2k:
All aggregated APIs face a finalization problem.  This will allow us to re-use the logic in the authentication operator.

/assign @p0lyn0mial @sanchezl @sohankunkerkar

### DIFF
--- a/pkg/operator/nsfinalizercontroller/finalizer_controller.go
+++ b/pkg/operator/nsfinalizercontroller/finalizer_controller.go
@@ -1,14 +1,11 @@
-package operator
+package nsfinalizercontroller
 
 import (
 	"fmt"
 	"reflect"
 	"time"
 
-	"github.com/openshift/cluster-openshift-apiserver-operator/pkg/operator/operatorclient"
-	"k8s.io/client-go/kubernetes/typed/core/v1"
-	"k8s.io/klog"
-
+	"github.com/openshift/library-go/pkg/operator/events"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -16,15 +13,18 @@ import (
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/wait"
 	kubeinformers "k8s.io/client-go/informers"
+	"k8s.io/client-go/kubernetes/typed/core/v1"
 	appsv1lister "k8s.io/client-go/listers/apps/v1"
 	corev1listers "k8s.io/client-go/listers/core/v1"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/util/workqueue"
-
-	"github.com/openshift/library-go/pkg/operator/events"
+	"k8s.io/klog"
 )
 
 type finalizerController struct {
+	name          string
+	namespaceName string
+
 	namespaceGetter v1.NamespacesGetter
 	podLister       corev1listers.PodLister
 	dsLister        appsv1lister.DaemonSetLister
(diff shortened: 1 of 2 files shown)


p0lyn0mial:
it doesn't compile
```
pkg/operator/nsfinalizercontroller/finalizer_controller.go:8:2: must be imported as github.com/openshift/library-go/pkg/operator/events
```

weird . . .

pkg/operator/nsfinalizercontroller/finalizer_controller.go
```diff
@@ -68,7 +73,7 @@ func NewFinalizerController(
 }
 
 func (c finalizerController) sync() error {
```
p0lyn0mial: so this controller simply removes all finalizers except `FinalizerKubernetes` from a namespace?

deads2k: > so this controller simply removes all finalizers except `FinalizerKubernetes` from a namespace?

It removes only the kube finalizer.  The kube finalizer requires every aggregated apiserver be running. When you're removing an aggregated apiserver's namespace that will come back, you don't want to remove the apiservices or controllers will do bad things, so this allows the namespace to be removed and recreated.

pkg/operator/starter.go
```diff
@@ -5,38 +5,36 @@ import (
...
 	"github.com/openshift/library-go/pkg/operator/unsupportedconfigoverridescontroller"
 	"github.com/openshift/library-go/pkg/operator/v1helpers"
-
-	"github.com/openshift/cluster-openshift-apiserver-operator/pkg/operator/configobservation/configobservercontroller"
-	"github.com/openshift/cluster-openshift-apiserver-operator/pkg/operator/operatorclient"
-	prune "github.com/openshift/cluster-openshift-apiserver-operator/pkg/operator/prunecontroller"
-	"github.com/openshift/cluster-openshift-apiserver-operator/pkg/operator/resourcesynccontroller"
-	"github.com/openshift/cluster-openshift-apiserver-operator/pkg/operator/workloadcontroller"
-	"github.com/openshift/library-go/pkg/operator/staticpod/controller/revision"
+	"k8s.io/apimachinery/pkg/api/errors"
```
stlaz: Why throw all the packages on one pile instead of keeping system, external, our and in-project imports separated in logical blocks?

pkg/operator/nsfinalizercontroller/finalizer_controller.go
```diff
@@ -79,14 +84,22 @@ func (c finalizerController) sync() error {
...
-	pods, err := c.podLister.Pods(operatorclient.TargetNamespace).List(labels.Everything())
+	// allow one minute of grace for most things to terminate.
+	// TODO now that we have conditions, we may be able to check specific conditions
+	deletedMoreThanAMinute := ns.DeletionTimestamp.Time.Add(1 * time.Minute).Before(time.Now())
+	if !deletedMoreThanAMinute {
+		c.queue.AddAfter(c.namespaceName, 1*time.Minute)
+		return nil
+	}
+
+	pods, err := c.podLister.Pods(c.namespaceName).List(labels.Everything())
```
mfojtik: @deads2k can we narrow this list to something that only select apiserver pods?

pkg/operator/nsfinalizercontroller/finalizer_controller.go
```diff
@@ -79,14 +84,22 @@ func (c finalizerController) sync() error {
...
 	if err != nil {
 		return err
 	}
 	if len(pods) > 0 {
 		return nil
 	}
-	dses, err := c.dsLister.DaemonSets(operatorclient.TargetNamespace).List(labels.Everything())
+	dses, err := c.dsLister.DaemonSets(c.namespaceName).List(labels.Everything())
```
mfojtik: do we expect more daemonsets in this namespace?

pkg/operator/starter.go
deads2k: > Why throw all the packages on one pile instead of keeping system, external, our and in-project imports separated in logical blocks? I know I asked this before, but this still feels counter-intuitive.

1. the tooling does a better job of keeping it in order
2. this list is still sorted
3. having separate blocks doesn't provide a practical benefit for tooling or hand inspection since they are alpha sorted.

pkg/operator/nsfinalizercontroller/finalizer_controller.go
deads2k: > do we expect more daemonsets in this namespace?

we know we create them for apiservers today.

pkg/operator/nsfinalizercontroller/finalizer_controller.go
deads2k: > @deads2k can we narrow this list to something that only select apiserver pods?

the point of this block is to ensure that all the pods are gone.  Scoping it down defeats the purpose.

pkg/operator/starter.go
stlaz: The tooling does the alpha sorting fine with multiple blocks, too.

I can see how having separate blocks could help expose certain copy-paste faster during development, it did for me in the past anyway.

I don't see how a single block provides a practical benefit for tooling or hand inspection either.

It was my impression that it was a common good practice that I've seen in multiple languages - C, Python, Haskell, Go, namely.

Being the only one who sees it like that probably makes me wrong though, I won't force this habit on projects that I do not own no more.

PR #277, opened 2019-12-10
Bug 1781574: Fix panic in workload controller

sttts:
Introduced through https://github.com/openshift/cluster-openshift-apiserver-operator/pull/248


stlaz:
/bugzilla refresh
what a fun bot stack trace we've got here, @stevekuznetsov this looks like a BZ issue, isn't it? There was some maintenance yesterday, maybe related?

PR #284, opened 2019-12-18
DEVEXP-424: Mounting node pull secrets

ricardomaraschini:
OpenShift API Server needs to make use of node pull credentials during `image stream import`. This PR mounts them inside the operand pods. For further details please refer to the [enhancement proposal](https://github.com/openshift/enhancements/pull/136).

Note: Please review https://github.com/openshift/openshift-apiserver/pull/83 as well

### DIFF
--- a/bindata/v3.11.0/openshift-apiserver/deploy.yaml
+++ b/bindata/v3.11.0/openshift-apiserver/deploy.yaml
@@ -65,6 +65,9 @@ spec:
         ports:
         - containerPort: 8443
         volumeMounts:
+        - mountPath: /var/lib/kubelet/config.json
+          name: node-pullsecrets
+          readOnly: true
         - mountPath: /var/run/configmaps/config
           name: config
         - mountPath: /var/run/secrets/etcd-client
@@ -95,6 +98,10 @@ spec:
             path: healthz
       terminationGracePeriodSeconds: 70 # a bit more than the 60 seconds timeout of non-long-running requests
       volumes:
+      - name: node-pullsecrets
+        hostPath:
+          path: /var/lib/kubelet/config.json
+          type: File
       - name: config
         configMap:
           name: config
--- a/pkg/operator/v311_00_assets/bindata.go
+++ b/pkg/operator/v311_00_assets/bindata.go
@@ -240,6 +240,9 @@ spec:
         ports:
         - containerPort: 8443
         volumeMounts:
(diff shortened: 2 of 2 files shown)


dmage:
Support of these pull secrets will be added to the API server by https://github.com/openshift/openshift-apiserver/pull/64.

I doubt about the prefix `/node`, I would remove it.

@mfojtik can you take a look?

ricardomaraschini:
> I doubt about the prefix `/node`, I would remove it.

I am also not happy with it tbh. What would be a good location in your opinion?
All mount points seems to use `/var/run/` as prefix, maybe `/var/run/node` ?

sttts:
/hold

until proposal is merged.

ricardomaraschini:
/unhold

enhancement proposal was merged.

bindata/v3.11.0/openshift-apiserver/deploy.yaml
```diff
@@ -95,6 +98,10 @@ spec:
             path: healthz
       terminationGracePeriodSeconds: 70 # a bit more than the 60 seconds timeout of non-long-running requests
       volumes:
+      - name: node-pullsecrets
+        hostPath:
+          path: /var/lib/kubelet/config.json
+          type: File
```
sttts: this is not optional. https://github.com/openshift/enhancements/pull/136/files#diff-a3521cf1ef50ae6c95913acbdca6106dR176 only states existence for worker nodes, but not master.

ricardomaraschini: Thanks for your comment @sttts. It made us chase the needed info down the road and yes, the pull secret file is always present on master nodes as well. The enhancement proposal got updated to include this missing information.

sttts: :+1:

mfojtik:
@adambkaplan @ricardomaraschini i assume the config.json is not something new that was added in last release and we don't have to solve a case where that file is missing, right?

Otherwise this change looks fine to me, the reloading should be handled by node reboot, which seems to be sufficient.

adambkaplan:
> i assume the config.json is not something new that was added in last release and we don't have to solve a case where that file is missing, right?

@mfojtik correct - this is the cluster pull secret that MCO (or something in that area) has been adding to the node for quite some time.

dmage:
@mfojtik can you tag it?

PR #285, opened 2020-01-03
switch to including hashes in annotation instead of ordered plumbing

deads2k:
This allows us to avoid requiring the same loops manage daemonsets and the resources they depend upon.  It also cleans up an SA dependency that wasn't logically required.

/assign @stlaz

### DIFF
--- a/pkg/operator/workloadcontroller/workload_controller_openshiftapiserver_v311_00.go
+++ b/pkg/operator/workloadcontroller/workload_controller_openshiftapiserver_v311_00.go
@@ -19,7 +19,6 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
-	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/scheme"
 	appsclientv1 "k8s.io/client-go/kubernetes/typed/apps/v1"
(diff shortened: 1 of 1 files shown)


pkg/operator/workloadcontroller/workload_controller_openshiftapiserver_v311_00.go
```diff
@@ -505,7 +480,26 @@ func manageOpenShiftAPIServerDaemonSet_v311_00_to_latest(
...
+		kubeClient,
+		resourcehash.NewObjectRef().ForConfigMap().InNamespace(operatorclient.TargetNamespace).Named("config"),
+		resourcehash.NewObjectRef().ForConfigMap().InNamespace(operatorclient.TargetNamespace).Named("image-import-ca"),
+	)
+	if err != nil {
+		return nil, false, fmt.Errorf("invalid dependency reference: %q", err)
+	}
+	inputHashes["desired.generation"] = fmt.Sprintf("%d", operatorConfig.ObjectMeta.Generation)
+	for k, v := range inputHashes {
+		annotationKey := fmt.Sprintf("operator.openshift.io/dep-%s", k)
```
deads2k: @stlaz do you see how a future iteration simply inverts this so that the manifest itself includes the deps that should be inspected and hashes built?

stlaz: You mean include them in "v3.11.0/openshift-apiserver/ds.yaml"? Could we do it right away?

deads2k: > You mean include them in "v3.11.0/openshift-apiserver/ds.yaml"? Could we do it right away?

Let's make the big cuts first so you have something reasonable to start following in the authentication-operator.

PR #286, opened 2020-01-03
replace verbosity based on string substitution for ease

deads2k:
This makes the verbosity easy for workloads to add in their operator manifest.


mfojtik:
(we need to do this in all operators...)

PR #287, opened 2020-01-03
vastly simplify the image registry trust configmap

deads2k:
Use the standard apply methods to gain proper eventing, caching, and diff computation.

/assign @stlaz @adambkaplan 

@stlaz @sttts @adambkaplan need a bug written for the direct dependency on the image-registry namespace above.  It should never have merged that way.  Communication between components happens through openshift-config-manged so we can see who is accessing what and avoid trip deployment deps.


stlaz:
/lgtm
Good change, the original code was almost unreadable compared to the new version.

adambkaplan:
@deads2k BZ filed: https://bugzilla.redhat.com/show_bug.cgi?id=1788235

PR #288, opened 2020-01-03
simplify logic to get the proxy configmap

deads2k:
There are already well established controllers to maintain the content of proxy configmaps.  This PR removes non-standard duplication of this logic. Auto-rollouts are still driven based on the hashes contained in the config configmap.

/assign @stlaz


stlaz:
/hold
this causes redeployment loop since the CM in question is being injected by the network operator

deads2k:
> /hold
> this causes redeployment loop since the CM in question is being injected by the network operator

That loop already logically exists, you just hid it behind more complex logic.  Since the network operator doesn't require the openshift-apiserver-operator, no logical cycle exists.

stlaz:
@deads2k What I meant is that there's a logical loop where you create the CM, it gets injected with CAs by the network operator, then you remove the data by reapplying the CM with empty data.

This was the very reason why I did not use the logic from this PR when I did the CM sync originally.

See tail of https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_cluster-openshift-apiserver-operator/288/pull-ci-openshift-cluster-openshift-apiserver-operator-master-e2e-aws/1672/artifacts/e2e-aws/pods/openshift-apiserver-operator_openshift-apiserver-operator-55f5c4dc84-8lf9g_openshift-apiserver-operator.log

deads2k:
> @deads2k What I meant is that there's a logical loop where you create the CM, it gets injected with CAs by the network operator, then you remove the data by reapplying the CM with empty data.
> 
> This was the very reason why I did not use the logic from this PR when I did the CM sync originally.

So I see.  I guess at some point we decided to stomp and not merge.  Guess I'll add a merge directive

deads2k:
/hold

needs upstream

@stlaz see if this fixes your concern.

stlaz:
@deads2k it's a hack which special-cases a single type of CM, on the other hand I can't think of anything better at the moment, I guess we'll have to go with it for the time being.

p0lyn0mial:
`e2e-aws-operator` will be fixed by https://github.com/openshift/cluster-openshift-apiserver-operator/pull/297

PR #289, opened 2020-01-06
merge hashing resources for forced DS rollouts

stlaz:
/assign @deads2k


deads2k:
/lgtm

can you add a test that twiddles these inputs and sees a redeployment happen?

/hold

for test

PR #290, opened 2020-01-07
split apiserver controllers initialization to a separate module

stlaz:
This is a crude stab at trying to separate and generalize setting up our own API servers.

/assign @deads2k

### DIFF
--- a/pkg/operator/apiservercontrollerset/apiservercontrollerset.go
+++ b/pkg/operator/apiservercontrollerset/apiservercontrollerset.go
@@ -1,0 +1,199 @@
+package apiservercontrollerset
+
+import (
+	"context"
+	"fmt"
+
+	configv1 "github.com/openshift/api/config/v1"
+	configv1client "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
+	configv1informers "github.com/openshift/client-go/config/informers/externalversions/config/v1"
+	"github.com/openshift/cluster-openshift-apiserver-operator/pkg/operator/apiservicecontroller"
+	"github.com/openshift/cluster-openshift-apiserver-operator/pkg/operator/nsfinalizercontroller"
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/loglevel"
+	"github.com/openshift/library-go/pkg/operator/status"
+	"github.com/openshift/library-go/pkg/operator/unsupportedconfigoverridescontroller"
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
+	"k8s.io/apimachinery/pkg/util/errors"
+	kubeinformers "k8s.io/client-go/informers"
+	"k8s.io/client-go/kubernetes"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+	apiregistrationv1 "k8s.io/kube-aggregator/pkg/apis/apiregistration/v1"
+	apiregistrationv1client "k8s.io/kube-aggregator/pkg/client/clientset_generated/clientset/typed/apiregistration/v1"
+	apiregistrationinformers "k8s.io/kube-aggregator/pkg/client/informers/externalversions"
+)
+
+type preparedAPIServerControllerSet struct {
+	controllers []controller
+}
+
+type controllerWrapper struct {
+	emptyAllowed bool
+	controller
+}
+
+type controller interface {
+	Run(ctx context.Context, workers int)
+}
+
+func (cw *controllerWrapper) prepare() (controller, error) {
+	if !cw.emptyAllowed && cw.controller == nil {
+		return nil, fmt.Errorf("missing controller")
+	}
+
+	return cw.controller, nil
+}
+
+// APIServerControllerSet is a set of controllers that maintain a deployment of
+// an API server and the namespace it's running in
+//
+// TODO: add workload and encryption controllers
+type APIServerControllerSet struct {
+	operatorClient v1helpers.OperatorClient
+	eventRecorder  events.Recorder
+
+	apiServiceController            controllerWrapper
+	clusterOperatorStatusController controllerWrapper
+	configUpgradableController      controllerWrapper
+	logLevelController              controllerWrapper
+	finalizerController             controllerWrapper
+}
+
+func NewAPIServerControllerSet(
+	operatorClient v1helpers.OperatorClient,
+	eventRecorder events.Recorder,
+) *APIServerControllerSet {
+	apiServerControllerSet := &APIServerControllerSet{
+		operatorClient: operatorClient,
+		eventRecorder:  eventRecorder,
+	}
+
+	return apiServerControllerSet
+}
+
+// WithConfigUpgradableController adds a controller for the operator to check for presence of
+// unsupported configuration and to set the Upgradable condition to false if it finds any
+func (cs *APIServerControllerSet) WithConfigUpgradableController() *APIServerControllerSet {
+	cs.configUpgradableController.controller = unsupportedconfigoverridescontroller.NewUnsupportedConfigOverridesController(cs.operatorClient, cs.eventRecorder)
+	return cs
+}
+
+func (cs *APIServerControllerSet) WithoutConfigUpgradableController() *APIServerControllerSet {
+	cs.configUpgradableController.controller = nil
+	cs.configUpgradableController.emptyAllowed = true
+	return cs
+}
+
+// WithLogLevelController adds a controller that configures logging for the operator
+func (cs *APIServerControllerSet) WithLogLevelController() *APIServerControllerSet {
+	cs.logLevelController.controller = loglevel.NewClusterOperatorLoggingController(cs.operatorClient, cs.eventRecorder)
+	return cs
+}
+
+func (cs *APIServerControllerSet) WithoutLogLevelController() *APIServerControllerSet {
+	cs.logLevelController.controller = nil
+	cs.logLevelController.emptyAllowed = true
+	return cs
+}
+
+func (cs *APIServerControllerSet) WithClusterOperatorStatusController(
+	clusterOperatorName string,
+	relatedObjects []configv1.ObjectReference,
+	clusterOperatorClient configv1client.ClusterOperatorsGetter,
+	clusterOperatorInformer configv1informers.ClusterOperatorInformer,
+	versionRecorder status.VersionGetter,
+) *APIServerControllerSet {
+	cs.clusterOperatorStatusController.controller = status.NewClusterOperatorStatusController(
+		clusterOperatorName,
+		relatedObjects,
+		clusterOperatorClient,
+		clusterOperatorInformer,
+		cs.operatorClient,
+		versionRecorder,
+		cs.eventRecorder,
+	)
+	return cs
+}
+
+func (cs *APIServerControllerSet) WithoutClusterOperatorStatusController() *APIServerControllerSet {
+	cs.clusterOperatorStatusController.controller = nil
+	cs.clusterOperatorStatusController.emptyAllowed = true
+	return cs
+}
+
+func (cs *APIServerControllerSet) WithAPIServiceController(
+	controllerName string,
+	apiServices []*apiregistrationv1.APIService,
+	apiregistrationInformers apiregistrationinformers.SharedInformerFactory,
+	apiregistrationv1Client apiregistrationv1client.ApiregistrationV1Interface,
+	kubeInformersForTargetNamesace kubeinformers.SharedInformerFactory,
+	kubeClient kubernetes.Interface,
+) *APIServerControllerSet {
+	cs.apiServiceController.controller = apiservicecontroller.NewAPIServiceController(
+		controllerName,
+		apiServices,
+		cs.operatorClient,
+		apiregistrationInformers,
+		apiregistrationv1Client,
+		kubeInformersForTargetNamesace,
+		kubeClient,
+		cs.eventRecorder,
+	)
+	return cs
+}
+
+func (cs *APIServerControllerSet) WithoutAPIServiceController() *APIServerControllerSet {
+	cs.apiServiceController.controller = nil
+	cs.apiServiceController.emptyAllowed = true
+	return cs
+}
+
+func (cs *APIServerControllerSet) WithFinalizerController(
+	targetNamespace string,
+	kubeInformersForTargetNamespace kubeinformers.SharedInformerFactory,
+	namespaceGetter corev1client.NamespacesGetter,
+) *APIServerControllerSet {
+	cs.finalizerController.controller = nsfinalizercontroller.NewFinalizerController(
+		targetNamespace,
+		kubeInformersForTargetNamespace,
+		namespaceGetter,
+		cs.eventRecorder,
+	)
+	return cs
+}
+
+func (cs *APIServerControllerSet) WithoutFinalizerController() *APIServerControllerSet {
+	cs.finalizerController.controller = nil
+	cs.finalizerController.emptyAllowed = true
+	return cs
+}
+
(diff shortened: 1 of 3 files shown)


pkg/operator/nsfinalizercontroller/finalizer_controller.go
```diff
@@ -13,15 +13,15 @@ import (
...
+	v1 "k8s.io/client-go/kubernetes/typed/core/v1"
 	appsv1lister "k8s.io/client-go/listers/apps/v1"
 	corev1listers "k8s.io/client-go/listers/core/v1"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/util/workqueue"
 	"k8s.io/klog"
 )
 
-type finalizerController struct {
+type FinalizerController struct {
```
deads2k: this change looks wrong

pkg/operator/apiservercontrollerset/apiservercontrollerset.go
```diff
@@ -0,0 +1,180 @@
...
+	apiServiceController  *apiservicecontroller.APIServiceController
+	configObserver        *configobservercontroller.ConfigObserver
+	encryptionControllers *encryption.Controllers
+	finalizerController   *nsfinalizercontroller.FinalizerController
+	pruneController       *prune.PruneController
+	revisionController    *revisioncontroller.RevisionController
+	workloadController    *workloadcontroller.OpenShiftAPIServerOperator
+}
+
+func NewAPIServerControllerSet(
```
deads2k: use a `With` pattern to allow pieces to be added/opt-in

deads2k:
start smaller with 
1. apiservice
2. finalizer
3. statuscontroller
4. loglevel
5. configupgradeable

pkg/operator/nsfinalizercontroller/finalizer_controller.go
stlaz: would
```
interface controller {
   Run(workers int, stopChan <-)
}

struct APIServerControllerSet {
    controllers []controller
    revisionController
    encryptionControllers
}
```
be better, in order to be able to run all the controllers in the `Run()` method of the struct?

stlaz:
I moved to the `apiservice` and `finalizer` being the base for every APIServer, the encryption controllers can be added using the `With()` pattern.

I am not sure about the statuscontroller, loglevel and configupgradable. Currently my understanding is that loglevel and status controllers should be set up only once per operator, and configupgradable once per a config resource.

If we are to plug the apiserver controller set among the other controllers of an existing operator, and if we are to use the `apiserver.config`, statuscontroller and loglevel should already be present for the operator, correct?

What I am not so sure about is the configupgradable, since the only resource that we currently have that we can use to set up other apiservers, is `openshiftapiserver.operator/cluster`, which is already handled by the openshift-apiserver operator.

It would make a lot of sense to have a new `oauthapiserver.operator/cluster` resource, is that the way we would like to go?

stlaz:
as for the second part, that was answered on slack, we'll be using the `*.operator` resource of the deploying operator, that's good.

pkg/operator/apiservercontrollerset/apiservercontrollerset.go
```diff
@@ -0,0 +1,179 @@
...
+	apiServiceController  *apiservicecontroller.APIServiceController
+	encryptionControllers *encryption.Controllers
+	finalizerController   *nsfinalizercontroller.FinalizerController
+	pruneController       *prune.PruneController
+	revisionController    *revisioncontroller.RevisionController
+
+	eventRecorder events.Recorder
+}
+
+func NewAPIServerControllerSet(
```
deads2k: I expected this to be nearly empty.

pkg/operator/apiservercontrollerset/apiservercontrollerset.go
```diff
@@ -0,0 +1,179 @@
...
+		targetNamespace,
+		kubeInformersForTargetNamesace,
+		kubeClient.CoreV1(),
+		apiServerControllerSet.eventRecorder,
+	)
+
+	return apiServerControllerSet
+}
+
+func (cs *APIServerControllerSet) WithEncryptionControllers(
```
deads2k: later PR

pkg/operator/apiservercontrollerset/apiservercontrollerset.go
```diff
@@ -0,0 +1,179 @@
...
+) *APIServerControllerSet {
+	apiServerControllerSet := &APIServerControllerSet{
+		kubeClient:              kubeClient,
+		operatorClient:          operatorClient,
+		apiserverOperatorClient: apiserverOperatorClient,
+
+		eventRecorder: eventRecorder,
+	}
+
+	apiServerControllerSet.apiServiceController = apiservicecontroller.NewAPIServiceController(
```
deads2k: add using with

pkg/operator/apiservercontrollerset/apiservercontrollerset.go
```diff
@@ -0,0 +1,179 @@
...
+		apiServices,
+		operatorClient,
+		apiregistrationInformers,
+		apiregistrationv1Client.ApiregistrationV1(),
+		kubeInformersForTargetNamesace,
+		kubeClient,
+		apiServerControllerSet.eventRecorder,
+	)
+
+	apiServerControllerSet.finalizerController = nsfinalizercontroller.NewFinalizerController(
```
deads2k: add using with

pkg/operator/apiservercontrollerset/apiservercontrollerset.go
```diff
@@ -0,0 +1,179 @@
...
+		[]string{"encryption-config-"},
+		cs.kubeClient.CoreV1(),
+		cs.kubeClient.CoreV1(),
+		kubeInformersForNamespaces,
+		cs.eventRecorder,
+	)
+
+	return cs
+}
+
```
deads2k: add a check method where you can return an error for an unhandled (no with or without) controller.

pkg/operator/apiservercontrollerset/apiservercontrollerset.go
```diff
@@ -0,0 +1,179 @@
...
+func (cs *APIServerControllerSet) Run(ctx context.Context) {
+	go cs.apiServiceController.Run(1, ctx.Done())
+	go cs.finalizerController.Run(1, ctx.Done())
```
deads2k: missing loglevel, status, configupgradeable.

deads2k:
> I moved to the `apiservice` and `finalizer` being the base for every APIServer, the encryption controllers can be added using the `With()` pattern.

We want to have a simple opt-in and opt-out, With and Without, so people know when options are there and think about them. This also lets you choose when to use this version of those control loops.

pkg/operator/apiservercontrollerset/dsnodeprovider.go
```diff
@@ -7,12 +7,11 @@ import (
 	"k8s.io/client-go/tools/cache"
 
 	encryptiondeployer "github.com/openshift/library-go/pkg/operator/encryption/deployer"
-
-	"github.com/openshift/cluster-openshift-apiserver-operator/pkg/operator/operatorclient"
 )
 
 // DaemonSetNodeProvider returns the node list from nodes matching the node selector of a DaemonSet
 type DaemonSetNodeProvider struct {
+	TargetNamespace                  string
```
deads2k: this one is for encryption?  move to an encryption package, not an apiserver one for now.

pkg/operator/nsfinalizercontroller/finalizer_controller.go
deads2k: ok, I can accept this for now.

pkg/operator/apiservercontrollerset/apiservercontrollerset.go
stlaz: I couldn't really decide whether to hide certain arguments that will otherwise repeat in the "With" calls, I'm fine with just having a controllers slice here + have attributes for controllers that don't include `Run(workers int, stopChan <-chan struct{})` in their interface.

pkg/operator/apiservercontrollerset/apiservercontrollerset.go
stlaz: I can see what we're doing now and how far we're going with the modularity, I'll add those.

pkg/operator/apiservercontrollerset/apiservercontrollerset.go
stlaz: I sticked with having just the `operatorClient` and `eventsRecorder` appear in this struct, I expect that at least the former should be the same for all the controllers appearing here, and there is probably no reason to have different event recorders either.

stlaz:
/test atll
aaaand the tests got stuck

stlaz:
/test all
the tests got stuck...

stlaz:
It seems that the encryption test failures were not in fact caused by this PR, they appear in the others, too.

stlaz:
/test e2e-aws-operator
this should now pass

pkg/operator/apiservercontrollerset/apiservercontrollerset.go
```diff
@@ -0,0 +1,158 @@
...
+type APIServerControllerSet struct {
+	operatorClient v1helpers.OperatorClient
+	eventRecorder  events.Recorder
+
+	apiServiceController            *apiservicecontroller.APIServiceController
+	clusterOperatorStatusController *status.StatusSyncer
+	configUpgradableController      *unsupportedconfigoverridescontroller.UnsupportedConfigOverridesController
+	logLevelController              *loglevel.LogLevelController
+	finalizerController             *nsfinalizercontroller.FinalizerController
+	revisionController              *revisioncontroller.RevisionController
```
deads2k: remove

pkg/operator/apiservercontrollerset/apiservercontrollerset.go
deads2k: acceptable

pkg/operator/apiservercontrollerset/apiservercontrollerset.go
```diff
@@ -0,0 +1,158 @@
...
+}
+
+// WithLogLevelController adds a controller that configures logging for the operator
+func (cs *APIServerControllerSet) WithLogLevelController() *APIServerControllerSet {
+	cs.logLevelController = loglevel.NewClusterOperatorLoggingController(cs.operatorClient, cs.eventRecorder)
+	return cs
+}
+
+func (cs *APIServerControllerSet) WithoutLogLevelController() *APIServerControllerSet {
+	cs.logLevelController = nil
```
deads2k: you will later need to know if this is an explicit choice to provide an error when someone doesn't handle  a new contorller.

pkg/operator/apiservercontrollerset/apiservercontrollerset.go
deads2k: > add a check method where you can return an error for an unhandled (no with or without) controller.

Still needed.  Use the trick we used in k8s.io/apiserver for .PrepareRun().Run() with types to ensure someone has to call and see if there are new controllers.

pkg/operator/apiservercontrollerset/apiservercontrollerset.go
```diff
@@ -0,0 +1,158 @@
...
+}
+
+func (cs *APIServerControllerSet) WithoutFinalizerController() *APIServerControllerSet {
+	cs.finalizerController = nil
+	return cs
+}
+
+func (cs *APIServerControllerSet) Run(ctx context.Context) {
+	go cs.apiServiceController.Run(ctx)
+	go runController(cs.clusterOperatorStatusController, ctx, 1)
```
deads2k: Cast and run inline.

deads2k: > Cast and run inline.

or assert.

deads2k:
direct looks good.  need to remove the unnecessary controllers to start.  You should find a way to manage with and without more cleanly.

pkg/operator/apiservercontrollerset/apiservercontrollerset.go
sttts: we use it for encryption I think

pkg/operator/apiservercontrollerset/apiservercontrollerset.go
```diff
@@ -0,0 +1,158 @@
...
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
+	kubeinformers "k8s.io/client-go/informers"
+	"k8s.io/client-go/kubernetes"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+	apiregistrationv1 "k8s.io/kube-aggregator/pkg/apis/apiregistration/v1"
+	apiregistrationv1client "k8s.io/kube-aggregator/pkg/client/clientset_generated/clientset/typed/apiregistration/v1"
+	apiregistrationinformers "k8s.io/kube-aggregator/pkg/client/informers/externalversions"
+)
+
+type APIServerControllerSet struct {
```
sttts: go doc what this is

pkg/operator/apiservercontrollerset/apiservercontrollerset.go
```diff
@@ -0,0 +1,158 @@
...
+}
+
+// WithConfigUpgradableController adds a controller for the operator to check for presence of
+// unsupported configuration and to set the Upgradable condition to false if it finds any
+func (cs *APIServerControllerSet) WithConfigUpgradableController() *APIServerControllerSet {
+	cs.configUpgradableController = unsupportedconfigoverridescontroller.NewUnsupportedConfigOverridesController(cs.operatorClient, cs.eventRecorder)
+	return cs
+}
+
+func (cs *APIServerControllerSet) WithoutConfigUpgradableController() *APIServerControllerSet {
```
sttts: do we really need these Without funcs?

sttts: I see where this is going (https://github.com/openshift/cluster-openshift-apiserver-operator/pull/290/files#r365360059). Then we need a special value (e.g. the zero value for the controller structs) to mark `configUpgradableController` to be configured. Or maybe easier: add a string set of set up controllers and compare in the `PrepareRun`-like method (maybe called `Finalize`?).

pkg/operator/apiservercontrollerset/apiservercontrollerset.go
stlaz: I wrapped the controllers so that I can check if they are unintentionally unset. The nsFinalizer struct is once again unimported thanks to that.

pkg/operator/apiservercontrollerset/apiservercontrollerset.go
stlaz: removed `RevisionController`, I forgot about it before

pkg/operator/apiservercontrollerset/apiservercontrollerset.go
stlaz: done

pkg/operator/apiservercontrollerset/apiservercontrollerset.go
stlaz: this may now make just a little bit more sense

pkg/operator/starter.go
```diff
@@ -174,15 +159,29 @@ func RunOperator(ctx context.Context, controllerConfig *controllercmd.Controller
 		),
 		configClient.ConfigV1(),
 		configInformers.Config().V1().ClusterOperators(),
-		operatorClient,
 		versionRecorder,
+	).WithConfigUpgradableController().
+		WithLogLevelController()
```
sttts: call prepare here directly. No need for the extra `runnableAPIServerControllers` var.

sttts: fine in a follow-up

PR #294, opened 2020-01-10
OAuth related API Services can be managed/served by an external server/operator

p0lyn0mial:
Please see https://github.com/openshift/enhancements/blob/master/enhancements/authentication/separate-oauth-resources.md for details.


pkg/operator/apiservicecontroller/apiservice_controller.go
```diff
@@ -230,3 +249,84 @@ func (c *APIServiceController) eventHandler() cache.ResourceEventHandler {
...
+	}
+
+	operatorVersionMajorMinor := semver.Version{}
+	operatorVersionMajorMinor.Major = operatorVersion.Major
+	operatorVersionMajorMinor.Minor = operatorVersion.Minor
+	oauthOperatorMajorMinor := semver.Version{}
+	oauthOperatorMajorMinor.Major = oauthOperatorVersion.Major
+	oauthOperatorMajorMinor.Minor = oauthOperatorVersion.Minor
+
+	return operatorVersionMajorMinor.Compare(oauthOperatorVersion) == 0
```
p0lyn0mial: this will be slightly different in `4.3`

pkg/operator/apiservicecontroller/apiservice_controller.go
```diff
@@ -32,6 +40,7 @@ const (
 type APIServiceController struct {
 	name        string
 	apiServices []*apiregistrationv1.APIService
+	apiGroupsManagedByOAuthServer []schema.GroupVersion
```
sttts: would use more generic names. We will have another split eventually.

p0lyn0mial: like `apiGroupsManagedByExternalServer` ?

pkg/operator/apiservicecontroller/apiservice_controller.go
```diff
@@ -230,3 +249,84 @@ func (c *APIServiceController) eventHandler() cache.ResourceEventHandler {
...
+
+// TODO: change the name
+func (c *APIServiceController) filterAPIServices() []*apiregistrationv1.APIService {
+
+	oauthOperatorPrecondition := c.oauthOperatorAvailableAndVersionMatch()
+
+	filteredApiServices := []*apiregistrationv1.APIService{}
+
+	// the c.apiServices list is authoritative
+	// it can compete with oauth-server in case of errors but it could save us on downgrade (the current coao and an old cao)
```
p0lyn0mial: if we remove this code from `4.4` then we can hit an issue on downgrade (the current coao and an old cao)

pkg/operator/apiservicecontroller/apiservice_controller.go
```diff
@@ -230,3 +253,93 @@ func (c *APIServiceController) eventHandler() cache.ResourceEventHandler {
...
+	for _, apiGroupManagedByOAuthServer := range c.apiGroupsManagedByOAuthServer {
+		if apiService.Spec.Group == apiGroupManagedByOAuthServer.Group && apiService.Spec.Version == apiGroupManagedByOAuthServer.Version {
+			return true
+		}
+	}
+	return false
+}
+
+func (c *APIServiceController) isAPIServiceAnnotatedByExternalServer(apiService *apiregistrationv1.APIService) bool {
+	existingApiService, err := c.apiregistrationv1Client.APIServices().Get(apiService.Name, metav1.GetOptions{})
```
p0lyn0mial: it could be taken from a lister.

p0lyn0mial:
@deads2k I think that our plan is to remove this code and move it to `4.3`, I am right?

pkg/operator/apiservicecontroller/apiservice_controller.go
```diff
@@ -230,3 +255,93 @@ func (c *APIServiceController) eventHandler() cache.ResourceEventHandler {
 		DeleteFunc: func(obj interface{}) { c.queue.Add(workQueueKey) },
 	}
 }
+
+// TODO: change the name
```
deads2k: godoc please

pkg/operator/apiservicecontroller/apiservice_controller.go
```diff
@@ -230,3 +255,93 @@ func (c *APIServiceController) eventHandler() cache.ResourceEventHandler {
...
+	filteredApiServices := []*apiregistrationv1.APIService{}
+
+	// the c.apiServices list is authoritative
+	// it can compete with oauth-server in case of errors but it could save us on downgrade (the current coao and an old cao)
+	for _, apiService := range c.apiServices {
+		if c.isAPIServicePotentiallyManagedByExternalServer(apiService) && oauthOperatorPrecondition && c.isAPIServiceAnnotatedByExternalServer(apiService) {
+			// TODO: log that this resource is being skipped on behalf of oauth-server
+			continue
+		}
+		filteredApiServices = append(filteredApiServices, apiService)
```
deads2k: API

pkg/operator/apiservicecontroller/apiservice_controller.go
```diff
@@ -230,3 +255,93 @@ func (c *APIServiceController) eventHandler() cache.ResourceEventHandler {
...
+
+	oauthOperatorPrecondition := c.oauthOperatorAvailableAndVersionMatch()
+
+	filteredApiServices := []*apiregistrationv1.APIService{}
+
+	// the c.apiServices list is authoritative
+	// it can compete with oauth-server in case of errors but it could save us on downgrade (the current coao and an old cao)
+	for _, apiService := range c.apiServices {
+		if c.isAPIServicePotentiallyManagedByExternalServer(apiService) && oauthOperatorPrecondition && c.isAPIServiceAnnotatedByExternalServer(apiService) {
+			// TODO: log that this resource is being skipped on behalf of oauth-server
```
deads2k: only when the list has changed from last time. When the list changes, emit an event too.

pkg/operator/apiservicecontroller/apiservice_controller.go
```diff
@@ -230,3 +255,93 @@ func (c *APIServiceController) eventHandler() cache.ResourceEventHandler {
 		DeleteFunc: func(obj interface{}) { c.queue.Add(workQueueKey) },
 	}
 }
+
+// TODO: change the name
+func (c *APIServiceController) filterAPIServices() []*apiregistrationv1.APIService {
+
+	oauthOperatorPrecondition := c.oauthOperatorAvailableAndVersionMatch()
+
+	filteredApiServices := []*apiregistrationv1.APIService{}
```
deads2k: track this persistently on the struct so you can inform about deltas, not about the same things every time.

pkg/operator/apiservicecontroller/apiservice_controller.go
```diff
@@ -230,3 +255,93 @@ func (c *APIServiceController) eventHandler() cache.ResourceEventHandler {
 		DeleteFunc: func(obj interface{}) { c.queue.Add(workQueueKey) },
 	}
 }
+
+// TODO: change the name
+func (c *APIServiceController) filterAPIServices() []*apiregistrationv1.APIService {
+
+	oauthOperatorPrecondition := c.oauthOperatorAvailableAndVersionMatch()
```
deads2k: This needs to be a function you can pass in.  Make it a `WithSharedResponsibility`

pkg/operator/apiservicecontroller/apiservice_controller.go
```diff
@@ -230,3 +255,93 @@ func (c *APIServiceController) eventHandler() cache.ResourceEventHandler {
...
+			// TODO: log that this resource is being skipped on behalf of oauth-server
+			continue
+		}
+		filteredApiServices = append(filteredApiServices, apiService)
+	}
+
+	return filteredApiServices
+}
+
+func (c *APIServiceController) oauthOperatorAvailableAndVersionMatch() bool {
```
deads2k: this needs to be a function you pass in during construction.

pkg/operator/apiservicecontroller/apiservice_controller.go
```diff
@@ -230,3 +255,93 @@ func (c *APIServiceController) eventHandler() cache.ResourceEventHandler {
...
+			break
+		}
+	}
+	if len(oauthOperatorVersionString) == 0 || len(operatorVersionString) == 0 {
+		return false
+	}
+
+	oauthOperatorVersion, err := semver.Parse(oauthOperatorVersionString)
+	if err != nil {
+		// TODO: log the err
```
deads2k: nah, ripple the error all the way up.

deads2k: > nah, ripple the error all the way up.

Yeah, this won't work well even in CI builds.  those don't have real versions.

pkg/operator/apiservicecontroller/apiservice_controller.go
```diff
@@ -230,3 +255,93 @@ func (c *APIServiceController) eventHandler() cache.ResourceEventHandler {
...
+			continue
+		}
+		filteredApiServices = append(filteredApiServices, apiService)
+	}
+
+	return filteredApiServices
+}
+
+func (c *APIServiceController) oauthOperatorAvailableAndVersionMatch() bool {
+	oauthOperator, err := c.clusterOperatorLister.Get("authentication")
```
deads2k: get the authentication.operator.openshift.io resource instead and create a new status field on it that says, "I'm managing this" then backport it to 4.3 and set it to false.

pkg/operator/apiservicecontroller/apiservice_controller.go
```diff
@@ -230,3 +255,93 @@ func (c *APIServiceController) eventHandler() cache.ResourceEventHandler {
...
+	}
+	if oauthOperator == nil {
+		return false
+	}
+
+	if !clusteroperatorv1helpers.IsStatusConditionTrue(oauthOperator.Status.Conditions, configv1.OperatorAvailable) {
+		return false
+	}
+
+	operatorVersionString := c.versionRecorder.GetVersions()["operator"]
```
deads2k: then all the version stuff is unnecessary.

pkg/operator/apiservicecontroller/apiservice_controller.go
```diff
@@ -6,12 +6,10 @@ import (
 	"strings"
 	"time"
 
-	operatorsv1 "github.com/openshift/api/operator/v1"
-	operatorv1 "github.com/openshift/api/operator/v1"
-	"github.com/openshift/library-go/pkg/operator/events"
-	"github.com/openshift/library-go/pkg/operator/resource/resourceapply"
-	"github.com/openshift/library-go/pkg/operator/status"
-	"github.com/openshift/library-go/pkg/operator/v1helpers"
```
deads2k: don't bother separating.  it's a waste of time.

pkg/operator/apiservicecontroller/apiservice_controller.go
p0lyn0mial: may I know why :) ?

pkg/operator/apiservicecontroller/apiservice_controller.go
p0lyn0mial: why ?

pkg/operator/apiservicecontroller/apiservice_controller.go
p0lyn0mial: good idea.

pkg/operator/apiservicecontroller/apiservice_controller.go
p0lyn0mial: `authentication.operator.openshift.io` doesn't provide a version information, the `KEP` explicitly states that we should compare the versions - https://github.com/deads2k/openshift-enhancements/blob/011cb767fb0e769f60f73af8dfb501847e458161/enhancements/authentication/separate-oauth-resources.md

pkg/operator/apiservicecontroller/apiservice_controller.go
deads2k: > why ?

This is going to move to library-go so we can re-use it in openshift and oauth server operators

pkg/operator/apiservicecontroller/apiservice_controller.go
deads2k: This is going to move to library-go so we can re-use it in openshift and oauth server operators

pkg/operator/apiservicecontroller/apiservice_controller.go
p0lyn0mial: won't returning an error put the operator into a degraded state?

pkg/operator/apiservicecontroller/apiservice_controller.go
deads2k: Yeah, turns out that doesn't work when compared against reality :(  .  Versions don't always exist.  We need a non-version dependent mechanism. Sorry, I didn't think about it until I saw it.

pkg/operator/apiservicecontroller/apiservice_controller.go
```diff
@@ -230,3 +261,106 @@ func (c *APIServiceController) eventHandler() cache.ResourceEventHandler {
...
+			return err
+		}
+
+		if !v1helpers.IsOperatorConditionTrue(operatorStatus.Conditions, operatorv1.OperatorStatusTypeAvailable) {
+			return fmt.Errorf("%s operator in %v is unavailable", authOperatorName, authOperatorGVR)
+		}
+
+		// let know the authentication operator that he will manage OAuth API Resources
+		if !v1helpers.IsOperatorConditionTrue(operatorStatus.Conditions, authOperatorConditionType) {
+			if _, _, updateError := v1helpers.UpdateStatus(authOperatorClient, v1helpers.UpdateConditionFn(operatorv1.OperatorCondition{
```
p0lyn0mial: `authentication` operator is not allowed to set this field.

pkg/operator/apiservicecontroller/apiservice_controller.go
```diff
@@ -230,3 +261,106 @@ func (c *APIServiceController) eventHandler() cache.ResourceEventHandler {
...
+	}
+
+	removed = oldSet.Difference(newSet).List()
+	added = newSet.Difference(oldSet).List()
+
+	return added, removed
+}
+
+// AuthenticationOperatorAvailableAndConditionSetPrecondition lets the authentication operator know that he will manage OAuth API Resources by setting the appropriate condition on its status
+func AuthenticationOperatorAvailableAndConditionSetPrecondition(authOperatorName string, authOperatorGVR schema.GroupVersionResource, authOperatorConditionType string, authOperatorClient v1helpers.OperatorClient) func() error {
```
p0lyn0mial: I will leave it here, for now, it can be easily moved outside.

pkg/operator/apiservicecontroller/apiservice_controller.go
```diff
@@ -230,3 +261,106 @@ func (c *APIServiceController) eventHandler() cache.ResourceEventHandler {
...
+		if !v1helpers.IsOperatorConditionTrue(operatorStatus.Conditions, operatorv1.OperatorStatusTypeAvailable) {
+			return fmt.Errorf("%s operator in %v is unavailable", authOperatorName, authOperatorGVR)
+		}
+
+		// let know the authentication operator that he will manage OAuth API Resources
+		if !v1helpers.IsOperatorConditionTrue(operatorStatus.Conditions, authOperatorConditionType) {
+			if _, _, updateError := v1helpers.UpdateStatus(authOperatorClient, v1helpers.UpdateConditionFn(operatorv1.OperatorCondition{
+				Type:    authOperatorConditionType,
+				Status:  operatorv1.ConditionTrue,
+				Reason:  "CASOSteppingDownOAuthAPIResourcesManagement",
```
p0lyn0mial: something tells me you won't like it.

pkg/operator/apiservicecontroller/apiservice_controller.go
```diff
@@ -230,3 +261,106 @@ func (c *APIServiceController) eventHandler() cache.ResourceEventHandler {
...
+		if c.isAPIServicePotentiallyManagedByExternalServer(apiService) && c.isAPIServiceAnnotatedByExternalServer(apiService) {
+			continue
+		}
+		filteredApiServices = append(filteredApiServices, apiService)
+	}
+
+	{
+		added, removed := diffAPIServices(c.managedAPIServices, filteredApiServices)
+		if len(added) > 0 {
+			c.eventRecorder.Eventf("OAuthAPIResourcesManagement", "The API Services for the following GVs %v will be managed by this operator (CAS-O)", added)
```
p0lyn0mial: This will have the following format `The API Services for the following GVs [v1.oauth.openshift.io v1.user.openshift.io] will be managed by an external operator/server`, do we want to make it pretty?

pkg/operator/apiservicecontroller/apiservice_controller.go
```diff
@@ -230,3 +261,106 @@ func (c *APIServiceController) eventHandler() cache.ResourceEventHandler {
...
+			klog.V(2).Infof("The API Services for the following GVs %v will be managed by this operator (CAS-O)", added)
+		}
+
+		if len(removed) > 0 {
+			c.eventRecorder.Eventf("OAuthAPIResourcesManagement", "The API Services for the following GVs %v will be managed by an external operator/server", removed)
```
p0lyn0mial: This will have the following format `The API Services for the following GVs [v1.oauth.openshift.io v1.user.openshift.io] will be managed by an external operator/server`, do we want to make it pretty?

pkg/operator/apiservicecontroller/apiservice_controller.go
```diff
@@ -230,3 +261,106 @@ func (c *APIServiceController) eventHandler() cache.ResourceEventHandler {
 		DeleteFunc: func(obj interface{}) { c.queue.Add(workQueueKey) },
 	}
 }
+
+// getAPIServicesToManage gets the desired list of API Services that will be managed by this operator
+// note that some services might be managed by an external operators/servers
+func (c *APIServiceController) getAPIServicesToManage() ([]*apiregistrationv1.APIService, error) {
```
deads2k: make this function the one to pass through?  Doing so allows simple things to be simple and makes hard things possible.

pkg/operator/apiservicecontroller/apiservice_controller.go
deads2k: > track this persistently on the struct so you can inform about deltas, not about the same things every time.

this still isn't persistent.

pkg/operator/apiservicecontroller/apiservice_controller.go
```diff
@@ -230,3 +261,106 @@ func (c *APIServiceController) eventHandler() cache.ResourceEventHandler {
...
+			c.eventRecorder.Eventf("OAuthAPIResourcesManagement", "The API Services for the following GVs %v will be managed by an external operator/server", removed)
+			klog.V(2).Infof("The API Services for the following GVs %v will be managed by an external operator/server", removed)
+		}
+	}
+
+	c.managedAPIServices = filteredApiServices
+	return c.managedAPIServices, nil
+}
+
+func (c *APIServiceController) isAPIServicePotentiallyManagedByExternalServer(apiService *apiregistrationv1.APIService) bool {
```
deads2k: this check looks extraneous

pkg/operator/apiservicecontroller/apiservice_controller.go
```diff
@@ -230,3 +261,106 @@ func (c *APIServiceController) eventHandler() cache.ResourceEventHandler {
...
+		if apiService.Spec.Group == apiGroupManagedByOAuthServer.Group && apiService.Spec.Version == apiGroupManagedByOAuthServer.Version {
+			return true
+		}
+	}
+	return false
+}
+
+func (c *APIServiceController) isAPIServiceAnnotatedByExternalServer(apiService *apiregistrationv1.APIService) bool {
+	existingApiService, err := c.apiregistrationv1Client.APIServices().Get(apiService.Name, metav1.GetOptions{})
+	if err != nil {
```
deads2k: not found needs different handling.  Assume ownership if not found.

p0lyn0mial: we will take over https://github.com/openshift/cluster-openshift-apiserver-operator/pull/294/files#diff-7c7cf45105d9733d46a21001a29c9a1eR42

pkg/operator/apiservicecontroller/apiservice_controller.go
```diff
@@ -230,3 +248,111 @@ func (c *APIServiceController) eventHandler() cache.ResourceEventHandler {
...
+	operatorPrecondition externalOperatorPreconditionFuncType,
+	authoritativeAPIServicesList []*apiregistrationv1.APIService,
+	eventRecorder events.Recorder,
+	apiGroupsManagedByExternalServer sets.String,
+	apiGroupsManagedByExternalServerAnnotation string) func() ([]*apiregistrationv1.APIService, error) {
+
+	observedAPIServices := authoritativeAPIServicesList
+	return func() ([]*apiregistrationv1.APIService, error) {
+		if externalOperatorPrecondition := operatorPrecondition(); externalOperatorPrecondition != nil {
+			return nil, externalOperatorPrecondition
```
p0lyn0mial: we could return `authoritativeAPIServicesList` when the operator is not available.

pkg/operator/apiservicecontroller/apiservice_controller.go
```diff
@@ -230,3 +248,111 @@ func (c *APIServiceController) eventHandler() cache.ResourceEventHandler {
...
+				klog.V(2).Infof("The API Services for the following GVs %v will be managed by an external operator/server", removed)
+			}
+		}
+
+		observedAPIServices = apiServicesToMange
+		return observedAPIServices, nil
+	}
+}
+
+func isAPIServicePotentiallyManagedByExternalServer(apiService *apiregistrationv1.APIService, apiGroupsManagedByExternalServer sets.String) bool {
```
p0lyn0mial: reminder for myself to remove it and use `apiGroupsManagedByExternalServer.Has`

pkg/operator/apiservicecontroller/apiservice_controller.go
deads2k: > we could return `authoritativeAPIServicesList` when the operator is not available.

yeah, that allows coasting.

pkg/operator/apiservicecontroller/apiservice_controller.go
```diff
@@ -230,3 +248,107 @@ func (c *APIServiceController) eventHandler() cache.ResourceEventHandler {
...
 	}
 }
+
+// externalOperatorPreconditionFunc represents a function that will be called by this operator
+// to determine if an external operator is ready to take over some API services
+type externalOperatorPreconditionFuncType func() error
+
+// GetAPIServicesToManageWrapper returns a function that returns the desired list of API Services that will be managed by this operator
+// note that some services might be managed by an external operators/servers
+func GetAPIServicesToManageWrapper(apiregistrationv1Client apiregistrationv1client.ApiregistrationV1Interface,
```
deads2k: I have a personal preference for a struct with functions rather than a closure.

pkg/operator/apiservicecontroller/apiservice_controller.go
```diff
@@ -230,3 +248,107 @@ func (c *APIServiceController) eventHandler() cache.ResourceEventHandler {
...
+	operatorPrecondition externalOperatorPreconditionFuncType,
+	authoritativeAPIServicesList []*apiregistrationv1.APIService,
```
deads2k: `allPossibleAPIServices`

pkg/operator/apiservicecontroller/apiservice_controller.go
```diff
@@ -230,3 +248,107 @@ func (c *APIServiceController) eventHandler() cache.ResourceEventHandler {
...
+	apiGroupsManagedByExternalServer sets.String,
+	apiGroupsManagedByExternalServerAnnotation string) func() ([]*apiregistrationv1.APIService, error) {
+
+	observedAPIServices := authoritativeAPIServicesList
+	return func() ([]*apiregistrationv1.APIService, error) {
+		if externalOperatorPrecondition := operatorPrecondition(); externalOperatorPrecondition != nil {
+			return nil, externalOperatorPrecondition
+		}
+
+		apiServicesToManage := []*apiregistrationv1.APIService{}
```
deads2k: newAPIServicesToManage

pkg/operator/apiservicecontroller/apiservice_controller.go
```diff
@@ -230,3 +248,107 @@ func (c *APIServiceController) eventHandler() cache.ResourceEventHandler {
...
+// GetAPIServicesToManageWrapper returns a function that returns the desired list of API Services that will be managed by this operator
+// note that some services might be managed by an external operators/servers
+func GetAPIServicesToManageWrapper(apiregistrationv1Client apiregistrationv1client.ApiregistrationV1Interface,
+	operatorPrecondition externalOperatorPreconditionFuncType,
+	authoritativeAPIServicesList []*apiregistrationv1.APIService,
+	eventRecorder events.Recorder,
+	apiGroupsManagedByExternalServer sets.String,
+	apiGroupsManagedByExternalServerAnnotation string) func() ([]*apiregistrationv1.APIService, error) {
+
+	observedAPIServices := authoritativeAPIServicesList
```
deads2k: currentAPIServicesToManage

pkg/operator/apiservicecontroller/apiservice_controller.go
```diff
@@ -230,3 +248,107 @@ func (c *APIServiceController) eventHandler() cache.ResourceEventHandler {
...
+		for _, apiService := range authoritativeAPIServicesList {
+			if apiGroupsManagedByExternalServer.Has(apiService.Name) && isAPIServiceAnnotatedByExternalServer(apiregistrationv1Client, apiService, apiGroupsManagedByExternalServerAnnotation) {
+				continue
+			}
+			apiServicesToManage = append(apiServicesToManage, apiService)
+		}
+
+		{
+			added, removed := diffAPIServices(observedAPIServices, apiServicesToManage)
+			if len(added) > 0 {
```
deads2k: if any changed, just write the entire new list.  I don't need separate message, but I only want messages for 

```
if len(added) >0 || len(removed) > 0{
   event("new list is %", list)
```

pkg/operator/apiservicecontroller/apiservice_controller.go
```diff
@@ -230,3 +248,107 @@ func (c *APIServiceController) eventHandler() cache.ResourceEventHandler {
...
+
+		observedAPIServices = apiServicesToManage
+		return observedAPIServices, nil
+	}
+}
+
+func isAPIServiceAnnotatedByExternalServer(apiregistrationv1Client apiregistrationv1client.ApiregistrationV1Interface, apiService *apiregistrationv1.APIService, apiGroupsManagedByExternalServerAnnotation string) bool {
+	existingApiService, err := apiregistrationv1Client.APIServices().Get(apiService.Name, metav1.GetOptions{})
+	if err != nil {
+		klog.V(4).Infof("unable to determine if the following API Service %s was annotated by an external operator due to %v", apiService.Name, err)
```
deads2k: event

pkg/operator/apiservicecontroller/apiservice_controller.go
```diff
@@ -230,3 +248,107 @@ func (c *APIServiceController) eventHandler() cache.ResourceEventHandler {
...
+			return err
+		}
+
+		if !v1helpers.IsOperatorConditionTrue(operatorStatus.Conditions, operatorv1.OperatorStatusTypeAvailable) {
+			return fmt.Errorf("%s operator in %v is unavailable", authOperatorName, authOperatorGVR)
+		}
+
+		// let know the authentication operator that he will manage OAuth API Resources
+		if !v1helpers.IsOperatorConditionTrue(operatorStatus.Conditions, authOperatorConditionType) {
+			if _, _, updateError := v1helpers.UpdateStatus(authOperatorClient, v1helpers.UpdateConditionFn(operatorv1.OperatorCondition{
```
deads2k: no, this should be a read operator on the authentication.operator.openshift.io.  the authentication operator sets the condition to true and this reads it.

pkg/operator/apiservicecontroller/apiservice_controller.go
```diff
@@ -230,3 +248,107 @@ func (c *APIServiceController) eventHandler() cache.ResourceEventHandler {
...
+		if err != nil {
+			return err
+		}
+
+		if !v1helpers.IsOperatorConditionTrue(operatorStatus.Conditions, operatorv1.OperatorStatusTypeAvailable) {
+			return fmt.Errorf("%s operator in %v is unavailable", authOperatorName, authOperatorGVR)
+		}
+
+		// let know the authentication operator that he will manage OAuth API Resources
+		if !v1helpers.IsOperatorConditionTrue(operatorStatus.Conditions, authOperatorConditionType) {
```
deads2k: I would still recommend an API field, but if @sttts is strongly against, I'll let you try this unstructured way first.  I don't think it makes the situation easier.

pkg/operator/apiservicecontroller/apiservice_controller.go
```diff
@@ -230,3 +248,107 @@ func (c *APIServiceController) eventHandler() cache.ResourceEventHandler {
...
+
+// AuthenticationOperatorAvailableAndConditionSetPrecondition lets the authentication operator know that he will manage OAuth API Resources by setting the appropriate condition on its status
+func AuthenticationOperatorAvailableAndConditionSetPrecondition(authOperatorName string, authOperatorGVR schema.GroupVersionResource, authOperatorConditionType string, authOperatorClient v1helpers.OperatorClient) func() error {
+	return func() error {
+		_, operatorStatus, _, err := authOperatorClient.GetOperatorState()
+		if err != nil {
+			return err
+		}
+
+		if !v1helpers.IsOperatorConditionTrue(operatorStatus.Conditions, operatorv1.OperatorStatusTypeAvailable) {
```
deads2k: I don't think available enters into it, but this is the spot to put `ManagingOAuthAPIServer`

pkg/operator/apiservicecontroller/apiservice_controller_test.go
```diff
@@ -153,24 +487,32 @@ func TestAvailableStatus(t *testing.T) {
...
-				},
-				precondition:            func() (bool, error) { return true, nil },
+				precondition:            func([]*apiregistrationv1.APIService) (bool, error) { return true, nil },
 				kubeClient:              kubeClient,
-				eventRecorder:           events.NewInMemoryRecorder(""),
+				eventRecorder:           eventRecorder,
 				operatorClient:          fakeOperatorClient,
 				apiregistrationv1Client: kubeAggregatorClient.ApiregistrationV1(),
 				versionRecorder:         status.NewVersionGetter(),
+				getAPIServicesToManageFunc: GetAPIServicesToManageWrapper(
```
deads2k: Yeah, a separate struct with a `NewAPIServicesToManage` function is likely to read easier.

pkg/operator/starter.go
```diff
@@ -131,9 +132,23 @@ func RunOperator(controllerConfig *controllercmd.ControllerContext) error {
...
 		"openshift-apiserver",
-		apiServices(),
+		versionRecorder,
+		apiservicecontroller.GetAPIServicesToManageWrapper(
+			apiregistrationv1Client.ApiregistrationV1(),
+			apiservicecontroller.AuthenticationOperatorAvailableAndConditionSetPrecondition("cluster", operatorv1.GroupVersion.WithResource("authentications"), "ManageOAuthAPIResources", authOperatorClient),
+			apiServices(),
+			controllerConfig.EventRecorder,
+			sets.NewString("v1.oauth.openshift.io", "v1.user.openshift.io"),
+			"cluster.authentication.operator.openshift.io/managed",
```
deads2k: `authentication.operator.openshift.io/managed`

pkg/operator/apiservicecontroller/apiservice_controller.go
p0lyn0mial: sure, if you don't like the condition I'm going to add a new field.

pkg/operator/apiservicecontroller/apiservice_controller.go
p0lyn0mial: oh, based on your previous comment I figured you want this operator to be authoritative.

pkg/operator/apiservicecontroller/apiservice_controller.go
p0lyn0mial: this is what you had in your KEP :)

p0lyn0mial:
@deads2k ready for final review.

pkg/operator/apiservicecontroller/apiservice_controller.go
```diff
@@ -48,7 +55,8 @@ type APIServiceController struct {
 
 func NewAPIServiceController(
 	name string,
-	apiServices []*apiregistrationv1.APIService,
+	versionRecorder status.VersionGetter,
+	getAPIServicesToManageFunc getAPIServicesToMangeFuncType,
```
deads2k: super-nit.  We generally call the type ***Func and instance **Fn

deads2k:
/lgtm
/approve

I recommend getting someone else to have a look too so they understand what's been done.

p0lyn0mial:
@deads2k please retag, I had to resolve some merge conflicts manually.

PR #297, opened 2020-01-14
bump(library-go) changes the next empty encryption key's name for encryption e2e test

p0lyn0mial:


vendor/github.com/openshift/library-go/test/library/encryption/helpers.go
```diff
@@ -143,11 +143,16 @@ func WaitForNextMigratedKey(t testing.TB, kubeClient kubernetes.Interface, prevK
...
 
-	t.Logf("Waiting up to %s for the next key %q, previous key was %q", waitPollTimeout.String(), nextKeyName, prevKeyMeta.Name)
+	t.Logf("Waiting up to %s for the next key %q, previous key was %q", waitPollTimeout.String(), nextKeyName, func(prevKeyName string) string {
+		if len(prevKeyName) == 0 {
+			return "no previous key"
+		}
+		return prevKeyName
+	}(prevKeyMeta.Name))
 	observedKeyName := prevKeyMeta.Name
 	if err := wait.Poll(waitPollInterval, waitPollTimeout, func() (bool, error) {
```
p0lyn0mial: if the controller is slow when creating the keys we can get into a state in which we expect the previous key to be `no previous key` which is wrong.

p0lyn0mial:
/hold

it looks promising, first, we need to merge https://github.com/openshift/library-go/pull/664

PR #298, opened 2020-01-14
Bug 1795693: change the next empty encryption key's name for encryption e2e test

p0lyn0mial:


p0lyn0mial:
quite big, not sure how can I make it smaller (update only library-go)

tnozicka:
> quite big, not sure how can I make it smaller (update only library-go)
you can bump only library-go (+transitives) with `go get -u github.com/openshift/library-go@master`

mfojtik:
@tnozicka i tried to switch to go.mod but @sttts stopped me (we don't really want to take any risks in this branch)

tnozicka:
It is fine, I like safe for backports. I was mentioning it only because it made my previous comment irrelevant.

PR #299, opened 2020-01-16
Add encryption controllers to apiservercontrollerset

stlaz:
This uses an unmerged version of the apiserver controller sets from library-go (https://github.com/openshift/library-go/pull/667), adds encryption controllers on top


p0lyn0mial:
/close

deprecated by https://github.com/openshift/cluster-openshift-apiserver-operator/pull/346

PR #300, opened 2020-01-17
etcdservicecontroller: observe etcd endpoints

mfojtik:
As openshift-apiserver depend on etcd endpoints being available for it, this observer will make openshift-apiserver degraded when they are missing.

### DIFF
--- a/pkg/operator/etcdservicecontroller/etcdservice_controller.go
+++ b/pkg/operator/etcdservicecontroller/etcdservice_controller.go
@@ -1,0 +1,186 @@
+package etcdservicecontroller
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/informers"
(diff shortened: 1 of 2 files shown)


pkg/operator/etcdservicecontroller/etcdservice_controller.go
```diff
@@ -0,0 +1,186 @@
...
+	c.cachesToSync = append(c.cachesToSync, kubeInformersForTargetNamespace.Core().V1().Endpoints().Informer().HasSynced)
+
+	return c
+}
+
+func (c *EtcdServiceController) sync() error {
+	foundConditions := []operatorv1.OperatorCondition{}
+
+	if _, err := c.servicesGetter.Services(c.targetNamespace).Get("etcd", metav1.GetOptions{}); err != nil {
+		foundConditions = append(foundConditions, operatorv1.OperatorCondition{
```
deads2k: I think I like the immediate "set it now" behavior in https://github.com/openshift/cluster-openshift-apiserver-operator/pull/301 .  What do you think in comparison?  it makes the pairs very obvious.

deads2k:
I prefer an immediate-set structure. Can you live with that?

The idea here is sound.

deads2k:
/close

superseded by https://github.com/openshift/cluster-openshift-apiserver-operator/pull/312

PR #302, opened 2020-01-17
bug 1792382: update conditions to look at daemonset status

deads2k:
/assign @mfojtik 

/hold

holding until we look at the conditions to see if they are correct.

### DIFF
--- a/pkg/operator/daemonsetstatuscontroller/daemonsetstatuscontroller.go
+++ b/pkg/operator/daemonsetstatuscontroller/daemonsetstatuscontroller.go
@@ -1,0 +1,226 @@
+package daemonsetstatuscontroller
+
+import (
+	"fmt"
+	"time"
+
+	operatorv1 "github.com/openshift/api/operator/v1"
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/informers"
+	appsv1listers "k8s.io/client-go/listers/apps/v1"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/util/workqueue"
+	"k8s.io/klog"
+)
+
+var (
+	controllerWorkQueueKey = "key"
+)
+
+type DaemonSetStatusController struct {
+	targetNamespace string
+	targetName      string
+
+	operatorClient  v1helpers.OperatorClient
+	daemonSetLister appsv1listers.DaemonSetLister
+
+	cachesToSync  []cache.InformerSynced
+	queue         workqueue.RateLimitingInterface
+	eventRecorder events.Recorder
+}
+
+// goes available=false when missing/no pods
+// goes progressing=true when rolling out
+// goes degraded=true when pod not available
+
+func NewDaemonSetStatusController(
+	targetNamespace string,
+	targetName string,
+	kubeInformersForTargetNamespace informers.SharedInformerFactory,
+	operatorClient v1helpers.OperatorClient,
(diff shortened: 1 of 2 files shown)


pkg/operator/daemonsetstatuscontroller/daemonsetstatuscontroller.go
```diff
@@ -0,0 +1,226 @@
...
+				Status:  operatorv1.ConditionTrue,
+				Reason:  "NewGeneration",
+				Message: fmt.Sprintf("daemonset/apiserver.openshift-operator: observed generation is %d, desired generation is %d.", actualDaemonSet.Status.ObservedGeneration, actualDaemonSet.ObjectMeta.Generation),
+			})))...,
+		)
+	} else {
+		errors = append(errors,
+			appendErrors(v1helpers.UpdateStatus(c.operatorClient, v1helpers.UpdateConditionFn(operatorv1.OperatorCondition{
+				Type:   "APIServerDaemonSetProgressing",
+				Status: operatorv1.ConditionFalse,
```
wking: I do not think this is sufficient for Progressing=False.  We should [also][1] look for `numberReady == currentNumberScheduled` and `updatedNumberScheduled == currentNumberScheduled`.

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1792382#c2

deads2k: > I do not think this is sufficient for Progressing=False. We should [also](https://bugzilla.redhat.com/show_bug.cgi?id=1792382#c2) look for `numberReady == currentNumberScheduled` and `updatedNumberScheduled == currentNumberScheduled`.

If daemonsets are tripping their generations right, then this is fine.  We aren't progressing.  If something is unavailable (don't use the ready condition), it will be degraded.

wking: > If something is unavailable (don't use the ready condition), it will be degraded.

If our desired state differs from our current state, I think we are attempting to reconcile the difference (e.g. by crashlooping the pod), and we are still attempting to progress when that is going on, even if we aren't actually making progress.  The [current API docs][1] aren't really clear on whether Progressing=True is for "state does not match our desired spec" or for the more narrow "... and we continue to make useful progress towards fixing that" (vs. "... and we just keep banging our heads into the same wall; no idea if we'll ever get there").  I'd thought it was just state != spec; you seem to feel it's "... but we're getting closer".  Can we land some doc language so it's clear to the CVO and other operator authors what the intended semantics are?

[1]: https://github.com/openshift/api/blob/e7ccdda6ba67b868baceb40b6db943355967f26d/config/v1/types_cluster_operator.go#L146-L148

deads2k: > clear on whether Progressing=True is for "state does not match our desired spec" or for the more narrow "... and we continue to make useful progress towards fixing that" (vs. "... and we just keep banging our heads into the same wall; no idea if we'll ever get there")

It is unreasonable to expect an operator author to be able to distinguish failures that will eventually resolve from those that won't.  There are a few narrow cases where it is possible and there are a multiplicity of unknown errors where it is not possible to know.

In this particular case, we have a tactical, narrow fix to alleviate pressure in a previous release and a far more complete change going into 4.4 that is too risky to backport: https://github.com/openshift/cluster-openshift-apiserver-operator/pull/303 .  I wouldn't devote more effort to the "edges of edges" handling for code that *will never exist in master*.

If the generation/status.generation of daemonsets doesn't work as expected, I'm open to more changes/niggling, but if it does, this *throw-away code* is good enough to alleviate our issues.

deads2k:
/hold

the status union isn't work right in 4.3.  See https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-openshift-apiserver-operator/302/pull-ci-openshift-cluster-openshift-apiserver-operator-release-4.3-e2e-aws/15 the diff between the high and low level status.

PR #304, opened 2020-01-17
make static resource controller you simply pass manifests to

deads2k:
/hold

has a fake bump in it

We use these in every operator we have.   It takes manifest and "just works"

```go
staticResourceController := staticresourcecontroller.NewStaticResourceController(
		"OpenshiftAPIServerStaticResources",
		v311_00_assets.Asset,
		[]string{
			"v3.11.0/openshift-apiserver/ns.yaml",
			"v3.11.0/openshift-apiserver/apiserver-clusterrolebinding.yaml",
			"v3.11.0/openshift-apiserver/svc.yaml",
			"v3.11.0/openshift-apiserver/sa.yaml",
			"v3.11.0/openshift-apiserver/trusted_ca_cm.yaml",
		},
		(&resourceapply.ClientHolder{}).WithKubernetes(kubeClient).WithAPIExtensionsClient(apiExtensionsClient),
		operatorClient,
		controllerConfig.EventRecorder,
	).
		AddEventHandler(kubeInformersForNamespaces.InformersFor(operatorclient.TargetNamespace).Core().V1().ConfigMaps().Informer().AddEventHandler).
		AddEventHandler(kubeInformersForNamespaces.InformersFor(operatorclient.TargetNamespace).Core().V1().ServiceAccounts().Informer().AddEventHandler).
		AddEventHandler(kubeInformersForNamespaces.InformersFor(operatorclient.TargetNamespace).Core().V1().Services().Informer().AddEventHandler).
		AddEventHandler(kubeInformersForNamespaces.InformersFor(operatorclient.TargetNamespace).Rbac().V1().ClusterRoleBindings().Informer().AddEventHandler).
		AddNamespaceEventHandler(kubeInformersForNamespaces.InformersFor(operatorclient.TargetNamespace).Core().V1().Namespaces().Informer().AddEventHandler, operatorclient.TargetNamespace)

```


pkg/operator/starter.go
```diff
@@ -81,6 +90,25 @@ func RunOperator(ctx context.Context, controllerConfig *controllercmd.Controller
...
+		},
+		(&resourceapply.ClientHolder{}).WithKubernetes(kubeClient).WithAPIExtensionsClient(apiExtensionsClient),
+		operatorClient,
+		controllerConfig.EventRecorder,
+	).
+		AddEventHandler(kubeInformersForNamespaces.InformersFor(operatorclient.TargetNamespace).Core().V1().ConfigMaps().Informer().AddEventHandler).
+		AddEventHandler(kubeInformersForNamespaces.InformersFor(operatorclient.TargetNamespace).Core().V1().ServiceAccounts().Informer().AddEventHandler).
+		AddEventHandler(kubeInformersForNamespaces.InformersFor(operatorclient.TargetNamespace).Core().V1().Services().Informer().AddEventHandler).
+		AddEventHandler(kubeInformersForNamespaces.InformersFor(operatorclient.TargetNamespace).Rbac().V1().ClusterRoleBindings().Informer().AddEventHandler).
+		AddNamespaceEventHandler(kubeInformersForNamespaces.InformersFor(operatorclient.TargetNamespace).Core().V1().Namespaces().Informer().AddEventHandler, operatorclient.TargetNamespace)
```
p0lyn0mial: why we are reacting to a namespace?

pkg/operator/starter.go
```diff
@@ -81,6 +90,25 @@ func RunOperator(ctx context.Context, controllerConfig *controllercmd.Controller
...
+		"OpenshiftAPIServerStaticResources",
+		v311_00_assets.Asset,
+		[]string{
+			"v3.11.0/openshift-apiserver/ns.yaml",
+			"v3.11.0/openshift-apiserver/apiserver-clusterrolebinding.yaml",
+			"v3.11.0/openshift-apiserver/svc.yaml",
+			"v3.11.0/openshift-apiserver/sa.yaml",
+			"v3.11.0/openshift-apiserver/trusted_ca_cm.yaml",
+		},
+		(&resourceapply.ClientHolder{}).WithKubernetes(kubeClient).WithAPIExtensionsClient(apiExtensionsClient),
```
p0lyn0mial: it was always like that or have they changed something in go 1.13? I think that taking the address of an anonymous obj didn't work.

pkg/operator/starter.go
```diff
@@ -234,6 +262,7 @@ func RunOperator(ctx context.Context, controllerConfig *controllercmd.Controller
 	configInformers.Start(ctx.Done())
 	dynamicInformers.Start(ctx.Done())
 
+	go staticResourceController.Run(ctx, 1)
```
p0lyn0mial: if it will ever make it to library-go it would be nice if we could let the callers know that setting workers will not be respected.

pkg/operator/staticresourcecontroller/static_resource_controller.go
```diff
@@ -0,0 +1,212 @@
...
+				Reason:  "SyncError",
+				Message: message,
+			})))...,
+		)
+	} else {
+		errors = append(errors,
+			appendErrors(v1helpers.UpdateStatus(c.operatorClient, v1helpers.UpdateConditionFn(operatorv1.OperatorCondition{
+				Type:   fmt.Sprintf("%sDegraded", c.name),
+				Status: operatorv1.ConditionFalse,
+				Reason: "AsExpected",
```
p0lyn0mial: we could clear the previous message.

pkg/operator/staticresourcecontroller/static_resource_controller.go
```diff
@@ -0,0 +1,212 @@
...
+			})))...,
+		)
+	}
+
+	return utilerrors.NewAggregate(errors)
+}
+
+func appendErrors(_ *operatorv1.OperatorStatus, _ bool, err error) []error {
```
p0lyn0mial: this method doesn't append errors, it converts an `err` to `[]error` :P

pkg/operator/starter.go
deads2k: > why we are reacting to a namespace?

when the namespace is created or deleted, we want to recheck to see which resources need to have work done.

pkg/operator/starter.go
deads2k: > it was always like that or have they changed something in go 1.13? I think that taking the address of an anonymous obj didn't work.

Uh, I just added this.  It seemed to compile

pkg/operator/starter.go
deads2k: > if it will ever make it to library-go it would be nice if we could let the callers know that setting workers will not be respected.

sure

pkg/operator/staticresourcecontroller/static_resource_controller.go
deads2k: > we could clear the previous message.

I'll read the linked code, but I'm pretty sure this is a newly constructed condition with a zero length message.

pkg/operator/staticresourcecontroller/static_resource_controller.go
deads2k: > this method doesn't append errors, it converts an `err` to `[]error` :P

sure, I'll rename.

deads2k:
/hold cancel

now based on the library-go impl

pkg/operator/workloadcontroller/workload_controller_openshiftapiserver_v311_00.go
```diff
@@ -44,20 +44,6 @@ func syncOpenShiftAPIServer_v311_00_to_latest(c OpenShiftAPIServerOperator, orig
 	errors := []error{}
 	operatorConfig := originalOperatorConfig.DeepCopy()
 
-	directResourceResults := resourceapply.ApplyDirectly(c.kubeClient, c.eventRecorder, v311_00_assets.Asset,
```
p0lyn0mial: @deads2k so what's your recommendation? To embed it into `workload` controller?

Isn't that a bit weird? A `daemonset` in `workload` controller depends on `statics resources` created by an external process without receiving any feedback

deads2k: > @deads2k so what's your recommendation? To embed it into `workload` controller?
> 
> Isn't that a bit weird? A `daemonset` in `workload` controller depends on `statics resources` created by an external process without receiving any feedback

The workload controller already depends on many loops like resource sync and cert rotation that are not embedded.  It expresses these dependencies in its resource.

PR #309, opened 2020-02-05
Bug 1801154: manifests/: Throw away unused high cardinality apiserver duration buckets

LiliC:
This shouldn't have any effect on our alerts or dashboards as the buckets are cumulative and for those purposes the remaining buckets should be sufficient.

Note: this change was made in "upstream" kube-promethues repo, see background for this coreos/kube-prometheus#387

Same as https://github.com/openshift/cluster-kube-apiserver-operator/pull/752


LiliC:
/assign @sttts

:)

LiliC:
cc @soltysh PTAL, thanks!

PR #313, opened 2020-02-20
bug 1804717: Switch to managing openshift-apiserver pods with a deployment

marun:
The switch from using a daemonset to using a deployment is intended to improve reliability during upgrade.

bz: https://bugzilla.redhat.com/show_bug.cgi?id=1804717

/cc @sttts @deads2k

### DIFF
--- a/bindata/v3.11.0/openshift-apiserver/deploy.yaml
+++ b/bindata/v3.11.0/openshift-apiserver/deploy.yaml
@@ -1,5 +1,5 @@
 apiVersion: apps/v1
-kind: DaemonSet
+kind: Deployment
 metadata:
   namespace: openshift-apiserver
   name: apiserver
@@ -7,17 +7,26 @@ metadata:
     app: openshift-apiserver
     apiserver: "true"
 spec:
-  updateStrategy:
-    type: RollingUpdate
+  replicas: 3
   selector:
     matchLabels:
-      app: openshift-apiserver
+      # Need to vary the app label from that used by the legacy
+      # daemonset ('openshift-apiserver') to avoid the legacy
+      # daemonset and its replacement deployment trying to try to
+      # manage the same pods.
+      #
+      # It's also necessary to use different labeling to ensure, via
+      # anti-affinity, at most one deployment-managed pod on each
+      # master node. Without label differentiation, anti-affinity
+      # would prevent a deployment-managed pod from running on a node
+      # that was already running a daemonset-managed pod.
+      app: openshift-apiserver-a
       apiserver: "true"
   template:
     metadata:
       name: openshift-apiserver
       labels:
-        app: openshift-apiserver
+        app: openshift-apiserver-a
         apiserver: "true"
     spec:
       serviceAccountName: openshift-apiserver-sa
@@ -119,4 +128,20 @@ spec:
       nodeSelector:
         node-role.kubernetes.io/master: ""
       tolerations:
-      - operator: Exists
+        # Ensure pod can be scheduled on master nodes
+      - key: "node-role.kubernetes.io/master"
+        operator: "Exists"
+        effect: "NoSchedule"
+        # Ensure pod can be evicted if the node is unreachable
+      - key: "node.kubernetes.io/unreachable"
+        operator: "Exists"
+        effect: "NoExecute"
+        tolerationSeconds: 120
+        # Ensure scheduling is delayed until node readiness
+        # (i.e. network operator configures CNI on the node)
+      - key: "node.kubernetes.io/not-ready"
+        operator: "Exists"
+        effect: "NoExecute"
+        tolerationSeconds: 120
+      # Anti-affinity is configured in code due to the need to scope
+      # selection to the computed pod template.
--- a/pkg/operator/starter.go
+++ b/pkg/operator/starter.go
@@ -26,6 +26,7 @@ import (
 	"github.com/openshift/library-go/pkg/operator/encryption"
 	"github.com/openshift/library-go/pkg/operator/encryption/controllers/migrators"
 	encryptiondeployer "github.com/openshift/library-go/pkg/operator/encryption/deployer"
+	"github.com/openshift/library-go/pkg/operator/events"
 	"github.com/openshift/library-go/pkg/operator/genericoperatorclient"
 	"github.com/openshift/library-go/pkg/operator/resource/resourceapply"
 	"github.com/openshift/library-go/pkg/operator/revisioncontroller"
@@ -38,7 +39,9 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
+	"k8s.io/klog"
 	apiregistrationv1 "k8s.io/kube-aggregator/pkg/apis/apiregistration/v1"
 	apiregistrationclient "k8s.io/kube-aggregator/pkg/client/clientset_generated/clientset"
 	apiregistrationinformers "k8s.io/kube-aggregator/pkg/client/informers/externalversions"
@@ -199,9 +202,9 @@ func RunOperator(ctx context.Context, controllerConfig *controllercmd.Controller
 		controllerConfig.EventRecorder,
 	)
 
-	nodeProvider := DaemonSetNodeProvider{
-		TargetNamespaceDaemonSetInformer: kubeInformersForNamespaces.InformersFor(operatorclient.TargetNamespace).Apps().V1().DaemonSets(),
-		NodeInformer:                     kubeInformersForNamespaces.InformersFor("").Core().V1().Nodes(),
+	nodeProvider := DeploymentNodeProvider{
+		TargetNamespaceDeploymentInformer: kubeInformersForNamespaces.InformersFor(operatorclient.TargetNamespace).Apps().V1().Deployments(),
+		NodeInformer:                      kubeInformersForNamespaces.InformersFor("").Core().V1().Nodes(),
 	}
 
 	deployer, err := encryptiondeployer.NewRevisionLabelPodDeployer("revision", operatorclient.TargetNamespace, kubeInformersForNamespaces, resourceSyncController, kubeClient.CoreV1(), kubeClient.CoreV1(), nodeProvider)
@@ -246,6 +249,10 @@ func RunOperator(ctx context.Context, controllerConfig *controllercmd.Controller
 			"Progressing",
 			// in 4.1.0-4.3.0 this was used for indicating the apiserver daemonset was available
 			"Available",
+			// in 4.1.0-4.3.z this was used for indicating the conditions of the apiserver daemonset.
+			"APIServerDaemonSetAvailable",
+			"APIServerDaemonSetProgressing",
+			"APIServerDaemonSetDegraded",
 		},
 		operatorClient,
 		controllerConfig.EventRecorder,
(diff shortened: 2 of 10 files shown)


marun:
@deads2k Some questions:

 - should removal of the daemonset block creation of the deployment (and vise-versa for cleanup in previous releases on downgrade)?
- the bz mentioned wiring HPA to enable scaling - should that be tackled in this PR or separately?

pkg/operator/starter.go
```diff
@@ -255,6 +261,18 @@ func RunOperator(ctx context.Context, controllerConfig *controllercmd.Controller
 		controllerConfig.Server.Handler.NonGoRestfulMux.Handle("/debug/controllers/resourcesync", debugHandler)
 	}
 
+	// Ensure removal of the daemonset that was previously used to manage apiserver pods.
+	dsClient := kubeClient.AppsV1().DaemonSets(operatorclient.TargetNamespace)
+	// TODO(marun) Should deletion block?
+	go wait.PollImmediateUntil(time.Minute, func() (bool, error) {
```
sttts: does this mean we unconditionally remove the daemonset before we know that the deployment is up already? We might want to check that at least one pod of the deployment is healthy.

pkg/operator/workloadcontroller/workload_controller_openshiftapiserver_v311_00.go
```diff
@@ -181,38 +181,43 @@ func syncOpenShiftAPIServer_v311_00_to_latest(c OpenShiftAPIServerOperator, orig
...
 			return nil
 		}))...,
 	)
 
-	daemonSetHasAllPodsAvailable := actualDaemonSet.Status.NumberAvailable == actualDaemonSet.Status.DesiredNumberScheduled
-	if !daemonSetHasAllPodsAvailable {
-		numNonAvailablePods := actualDaemonSet.Status.DesiredNumberScheduled - actualDaemonSet.Status.NumberAvailable
+	desiredReplicas := int32(1)
+	if actualDeployment.Spec.Replicas != nil {
+		desiredReplicas = *(actualDeployment.Spec.Replicas)
```
sttts: where do we set that number?

marun: Done. Set in the deployment yaml.

deads2k:
> * should removal of the daemonset block creation of the deployment (and vise-versa for cleanup in previous releases on downgrade)?

no, just rollout. We can tolerate having some extras for a bit.

pkg/operator/starter.go
marun: Done.

marun:
@sttts Updated, PTAL

bindata/v3.11.0/openshift-apiserver/deploy.yaml
```diff
@@ -120,3 +121,13 @@ spec:
         node-role.kubernetes.io/master: ""
       tolerations:
       - operator: Exists
+      # Ensure that no more than one apiserver pod is deployed per master
+      affinity:
+        podAntiAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+          - podAffinityTerm:
+              labelSelector:
+                matchLabels:
```
deads2k: on an upgrade, will this end up avoiding the existing daemonsets?

pkg/operator/starter.go
```diff
@@ -246,6 +248,10 @@ func RunOperator(ctx context.Context, controllerConfig *controllercmd.Controller
 			"Progressing",
 			// in 4.1.0-4.3.0 this was used for indicating the apiserver daemonset was available
 			"Available",
+			// in 4.4.0 this was used for indicating the conditions of the apiserver daemonset.
```
deads2k: 4.1.0-4.3.z

pkg/operator/starter.go
```diff
@@ -326,3 +334,58 @@ func apiServicesReferences() []configv1.ObjectReference {
...
+	// daemonset and deployment both use the same name
+	resourceName := "apiserver"
+
+	dsClient := kubeClient.AppsV1().DaemonSets(operatorclient.TargetNamespace)
+	deployClient := kubeClient.AppsV1().Deployments(operatorclient.TargetNamespace)
+
+	// It's safe not to check the error returned by the poll function so long
+	// as conditionFunc never returns an error. The poll function will only
+	// exit on success or if the stop channel halts execution.
+	go wait.PollImmediateUntil(time.Minute, func() (bool, error) {
```
deads2k: `UntilWithContext` please.  This should run forever.  I'm ok doing it on two cadences if you prefer (fast until we delete, slow after that), but this will run every time the pod restarts, which is just a controller on an unpredictable resync cadence.

bindata/v3.11.0/openshift-apiserver/deploy.yaml
marun: Done. Updated deployment label to use `app: openshift-apiserver-a` to ensure deployment and daemonset won't try to manage the same pods.

pkg/operator/starter.go
marun: Done. I didn't bother with 2 cadences, can address that if you feel strongly.

pkg/operator/starter.go
marun: Done

marun:
@deads2k Updated, PTAL

deads2k:
/lgtm

this could be improved by emitting some events, but we need this badly enough not to quibble.

marun:
@deads2k What events would you like to see? Successful deletion of daemonset? Deployment doesn't have replicas yet?

marun:
Ugh, for some reason the anti-affinity configuration wasn't seeing the value for topologyKey. I've moved it, hopefully that has a different result.

deads2k:
> Ugh, for some reason the anti-affinity configuration wasn't seeing the value for topologyKey. I've moved it, hopefully that has a different result.

doesn't seem like that should matter

deads2k:
`requiredDuringSchedulingIgnoredDuringExecution` doesn't have a weight and has no `podAffinityTerm`.  it's just straight `PodAffinityTerm` type

deads2k:
/lgtm cancel

Work out why this can no longer start

marun:
> Work out why this can no longer start

Anti-affinity keying on labels common to all replicasets was preventing rollout. The pod template is now labeled with a hash of the template that anti-affinity also selects for to resolve this, and ensure that a only single pod per template hash can be deployed to a node.

deads2k:
> Anti-affinity keying on labels common to all replicasets was preventing rollout. The pod template is now labeled with a hash of the template that anti-affinity also selects for, and this ensures that only a single pod per template hash can be deployed to a node.

Still not working.

Why not simply put a UUID in there?  We don't actually care what the value is for adoption, right?

marun:
> Why not simply put a UUID in there? We don't actually care what the value is for adoption, right?

Trying with a UUID. A content hash works fine in the naive case but is likely disjoint wrt how deployments manage rollouts with replicasets. Given that `ApplyDeployment` will only write when metadata changes or a generation change was observed maybe this will work (though potentially at the cost of slower rollouts)?

Switching from required to preferred anti-affinity would likely fix the rollout problem, but I imagine poor degenerative behavior e.g. if a master node went offline, an apiserver pod could be rescheduled to a different master that was already hosting an apiserver pod and it wouldn't be rescheduled one the master came back up.

marun:
Updated to use explicit tolerations which should result in e2e passing, PTAL.

bindata/v3.11.0/openshift-apiserver/deploy.yaml
```diff
@@ -1,23 +1,32 @@
...
 metadata:
   namespace: openshift-apiserver
   name: apiserver
   labels:
     app: openshift-apiserver
     apiserver: "true"
 spec:
-  updateStrategy:
-    type: RollingUpdate
+  replicas: 3
```
sttts: what does this mean for single-node clusters? Will we be always progressing?

pkg/operator/workloadcontroller/workload_controller_openshiftapiserver_v311_00.go
```diff
@@ -363,6 +369,7 @@ func manageOpenShiftAPIServerDaemonSet_v311_00_to_latest(
 
 	required.Labels["revision"] = strconv.Itoa(int(operatorConfig.Status.LatestAvailableRevision))
 	required.Spec.Template.Labels["revision"] = strconv.Itoa(int(operatorConfig.Status.LatestAvailableRevision))
+	required.Spec.Template.Labels["previousGeneration"] = strconv.Itoa(int(operatorConfig.Status.LatestAvailableRevision))
```
sttts: why this?

pkg/operator/workloadcontroller/workload_controller_openshiftapiserver_v311_00.go
```diff
@@ -181,38 +182,43 @@ func syncOpenShiftAPIServer_v311_00_to_latest(c OpenShiftAPIServerOperator, orig
...
 
-	daemonSetHasAllPodsAvailable := actualDaemonSet.Status.NumberAvailable == actualDaemonSet.Status.DesiredNumberScheduled
-	if !daemonSetHasAllPodsAvailable {
-		numNonAvailablePods := actualDaemonSet.Status.DesiredNumberScheduled - actualDaemonSet.Status.NumberAvailable
+	desiredReplicas := int32(1)
+	if actualDeployment.Spec.Replicas != nil {
+		desiredReplicas = *(actualDeployment.Spec.Replicas)
+	}
+
+	deploymentHasAllPodsAvailable := actualDeployment.Status.AvailableReplicas == desiredReplicas
```
sttts: do we enforce that two replicasets for the deployment don't overlap and increase this number temporarily although we are still progressing?

bindata/v3.11.0/openshift-apiserver/deploy.yaml
tnozicka: shouldn't that be observed from the number of masters in the cluster? if someone scales them to 5?

tnozicka:
> The switch from using a daemonset to using a deployment is intended to improve reliability during upgrade.

it would be nice if the description said `why`

bindata/v3.11.0/openshift-apiserver/deploy.yaml
sttts: 3 instances in a 5 master node cluster at least sounds sane and doesn't break.

marun:
> > The switch from using a daemonset to using a deployment is intended to improve reliability during upgrade.
> 
> it would be nice if the description said `why`

I'll leave it to @deads2k to explain. I have no idea why and the bz is short on details.

bindata/v3.11.0/openshift-apiserver/deploy.yaml
marun: > what does this mean for single-node clusters? Will we be always progressing?

Good catch. The logic should be determining health by checking whether the replica count is equal or greater to the number of masters to maintain the invariant previously guaranteed by the daemonset.

marun: > 3 instances in a 5 master node cluster at least sounds sane and doesn't break.

Is that preferable to adjusting the replica count according to the number of masters that are present?

pkg/operator/workloadcontroller/workload_controller_openshiftapiserver_v311_00.go
marun: Removed. I have no recollection of where that came from.

pkg/operator/workloadcontroller/workload_controller_openshiftapiserver_v311_00.go
marun: No, we do not enforce no overlap. As per your previous comment, health should be determined by comparing replica count to master count.

bindata/v3.11.0/openshift-apiserver/deploy.yaml
deads2k: single node clusters are not a supported platform.  Refinements could be made, but I don't see them as blocking us stablizing upgrades for supported platforms.

deads2k:
> > The switch from using a daemonset to using a deployment is intended to improve reliability during upgrade.
> 
> it would be nice if the description said `why`

There is an issue draining nodes.  Daemonsets are not evicted early.  When the termination finally happens, there is an issue where graceful deletion doesn't take place.  In addition, it is opening us up to SDN restart issues which result in API outages.

marun:
#324 addresses outstanding comments on this PR.

If the cherry-pick bot can't apply cleanly, does that imply that a manual PR is required?

PR #314, opened 2020-02-20
Bug 1805388: Cleanup 4.4 deployment on downgrade to 4.3

marun:
4.3 uses a daemonset to manage apiserver pods, so the 4.4 deployment is likely to cause trouble if not removed.

### DIFF
--- a/pkg/operator/starter.go
+++ b/pkg/operator/starter.go
@@ -8,8 +8,10 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/wait"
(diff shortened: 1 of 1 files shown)


pkg/operator/starter.go
```diff
@@ -229,3 +233,55 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
...
+		if deploy.ObjectMeta.DeletionTimestamp != nil {
+			// Done - deployment has been marked for deletion
+			return
+		}
+
+		// Check that the daemonset managing the apiserver pods has at last one available replica
+		ds, err := dsClient.Get(resourceName, metav1.GetOptions{})
+		if errors.IsNotFound(err) {
+			// No available replicas if the daemonset doesn't exist
+			return
```
sttts: same comment as for other PR: why not stop the loop?

marun: Please confer with @deads2k - this strategy is as per his requirement.

PR #319, opened 2020-02-21
waits for extension-apiserver-authentication before rolling out a new version

p0lyn0mial:


pkg/operator/workloadcontroller/workload_controller.go
```diff
@@ -142,6 +145,21 @@ func (c OpenShiftAPIServerOperator) sync() error {
 		return nil
 	}
 
+	// block until extension-apiserver-authentication configmap is available
+	// see https://bugzilla.redhat.com/show_bug.cgi?id=1795163#c19 to check why we have to wait for it
+	// in the future we need to change upstream code to be more dynamic
```
sttts:
```
// block until extension-apiserver-authentication configmap is available to avoid 
// that openshift-apiserver starts up with request header setting (which are not dynamically reloaded).
```

pkg/operator/workloadcontroller/workload_controller.go
```diff
@@ -142,6 +145,21 @@ func (c OpenShiftAPIServerOperator) sync() error {
...
+	if !c.haveObservedExtensionConfigMap {
+		_, err = c.kubeClient.CoreV1().ConfigMaps(metav1.NamespaceSystem).Get("extension-apiserver-authentication", metav1.GetOptions{})
+		if apierrors.IsNotFound(err) {
+			klog.Infof("Waiting for %q configmap in %q namespace to be available", "extension-apiserver-authentication", metav1.NamespaceSystem)
```
sttts: how do we wait? Are we triggered by that configmap here?

p0lyn0mial:
/close

as this is not entirely what we need.

PR #320, opened 2020-02-24
Bug 1806671: openshift-apiserver should never use the bootstrap etcd member

sanchezl:
oas should never use the bootstrap etcd member.
Switch to the new `host-etcd-2`, which never contains the etcd bootstrap member, endpoints resource to determine the storage urls.


deads2k:
1. current code fails on a not found.  
2. that constant is in the wrong spot

sanchezl:
Follow up PR for 4.5+: https://github.com/openshift/cluster-openshift-apiserver-operator/pull/321

PR #323, opened 2020-02-25
Bug 1806918: remove run-level=1 from openshift-apiserver NS

stlaz:
The apiserver DS runs with privileged SCC anyway.


stlaz:
/test e2e-aws-upgrade
suspicious, though, looks like no nodes are willing to reach the new generation

PR #324, opened 2020-02-27
bug 1804717: Cleanup switch to managing openshift-apiserver pods with a deployment

marun:
This PR addresses comments and otherwise unresolved issues from #313 

- Sets the replica count according to the number of master nodes
- Ensures the operator doesn't report degraded if the number of available nodes exceeds the desired replica count (a likely scenario during rollout)
- Removes the unintentionally applied `previousGeneration` label from the deployment
- Updates previously unnoticed daemonset-derived variable names.

cc: @deads2k @sttts @tnozicka

### DIFF
--- a/pkg/operator/workloadcontroller/workload_controller.go
+++ b/pkg/operator/workloadcontroller/workload_controller.go
@@ -14,10 +14,12 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/wait"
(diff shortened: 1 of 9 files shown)


pkg/operator/workloadcontroller/workload_controller.go
```diff
@@ -85,6 +94,17 @@ func NewWorkloadController(
 		queue: workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "OpenShiftAPIServerOperator"),
 	}
 
+	// This function supports determining the number of master nodes
+	// to allow setting the deployment replica count to match.
+	c.countNodes = func(nodeSelector map[string]string) (*int32, error) {
```
deads2k: Too big for an anonymous function. Make this an actual function on the `OpenshfitAPIServerOperator`. Then assign this function to `countNodes` here.

deads2k:
/lgtm
/hold

hold for the factorization update.

pkg/operator/workloadcontroller/workload_controller.go
marun: Done

deads2k:
/hold cancel
/lgtm

if this causes multiple rollouts, we may have to reconsider, but it seems like a reasonable way to be flexible for now.

PR #326, opened 2020-02-28
[release-4.3] Bug 1809532: waits for requestheader-client-ca-file in extension-apiserver-authentication  before rolling out a new version

p0lyn0mial:


p0lyn0mial:
this is the version Praveen and I verified against `4.3` cluster.

PR #330, opened 2020-03-05
Bug 1806918: remove run-level=1 from openshift-apiserver-operator NS

sttts:


stlaz:
/hold
the upgraded NS still has the run-level label

stlaz:
/retest
/hold cancel
it's still better than what we've got now

PR #334, opened 2020-03-11
Bug 1812583: Normalize CPU requests on masters

smarterclayton:
The openshift-apiserver uses approximately 10% of master CPU in a
reasonable medium sized workload. Given a 1 core per master baseline
(since CPU is compressible and shared), assign the openshift-apiserver
roughly 10% of that core on each master.

4.4 bug is 1812709


smarterclayton:
@sttts poke

PR #339, opened 2020-03-18
Bug 1814621: storage urls observer: prevent event/log spam

stlaz:
DeepEqual fails on type comparison resulting in spamming update events


mfojtik:
This is worth fixing and we don't have to wait for QE to verify that this is gone, we know it is :-)

PR #340, opened 2020-03-18
Bug 1814692 - openshift-apiserver should create stable backend

sanchezl:
Excessive events like this:
```
Mar 14 04:38:32.591 I ns/openshift-apiserver-operator deployment/openshift-apiserver-operator Updated storage urls to https://10.0.0.3:2379,https://10.0.0.4:2379,https://10.0.0.5:2379 (3001 times)
```

### DIFF
--- a/pkg/operator/configobservation/etcdobserver/observe_etcd.go
+++ b/pkg/operator/configobservation/etcdobserver/observe_etcd.go
@@ -39,7 +39,7 @@ func ObserveStorageURLs(genericListers configobserver.Listers, recorder events.R
 
 	observedConfig := map[string]interface{}{}
 
-	var etcdURLs sort.StringSlice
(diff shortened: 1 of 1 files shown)


pkg/operator/configobservation/etcdobserver/observe_etcd.go
```diff
@@ -87,6 +87,7 @@ func ObserveStorageURLs(genericListers configobserver.Listers, recorder events.R
 		errs = append(errs, err)
 	}
 
+	sort.Strings(currentEtcdURLs)
```
sttts: this line is not necessary I think. But doesn't harm either.

stlaz:
/close
fixed in https://github.com/openshift/cluster-openshift-apiserver-operator/pull/338

PR #344, opened 2020-03-25
OAuthAPIServerController

p0lyn0mial:
this PR adds `OAuthAPIServerController` that will manage `encryption-config-openshift-oauth-apiserver` as described in https://github.com/openshift/enhancements/blob/master/enhancements/etcd/etcd-encryption-for-separate-oauth-apis.md

### DIFF
--- a/pkg/operator/oauthapiencryptioncontroller/oauthapiencryptioncontroller.go
+++ b/pkg/operator/oauthapiencryptioncontroller/oauthapiencryptioncontroller.go
@@ -1,0 +1,133 @@
+package oauthapiencryptioncontroller
+
+import (
+	"context"
+	"fmt"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/openshift/cluster-openshift-apiserver-operator/pkg/operator/operatorclient"
+	"github.com/openshift/library-go/pkg/controller/factory"
+	"github.com/openshift/library-go/pkg/operator/encryption/encryptionconfig"
+	encryptionsecret "github.com/openshift/library-go/pkg/operator/encryption/secrets"
+	encryptionstate "github.com/openshift/library-go/pkg/operator/encryption/state"
+	"github.com/openshift/library-go/pkg/operator/events"
+	operatorv1helpers "github.com/openshift/library-go/pkg/operator/v1helpers"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+	corev1listers "k8s.io/client-go/listers/core/v1"
+)
+
+const (
+	encryptionConfigManagedBy      = "encryption.apiserver.operator.openshift.io/managed-by"
+	encryptionConfigManagedByValue = `WARNING: DO NOT REMOVE.
+This annotation indicates that OAS-O manages this secret.`
+)
+
+type oauthAPIServerController struct {
+	targetNamespace               string
+	oauthAPIServerTargetNamespace string
+
+	secretLister corev1listers.SecretNamespaceLister
+	secretClient corev1client.SecretInterface
+}
+
+// New creates OAuthAPIServerController that will manage encryption-config-openshift-oauth-apiserver in openshift-config-managed namespace as described in https://github.com/openshift/enhancements/blob/master/enhancements/etcd/etcd-encryption-for-separate-oauth-apis.md
+// Note that this code will be removed in the future release (4.6)
+func New(
+	name string,
+	targetNamespace string,
+	oauthAPIServerTargetNamespace string,
+	secretClient corev1client.SecretsGetter,
+	kubeInformersForNamespaces operatorv1helpers.KubeInformersForNamespaces,
+	eventRecorder events.Recorder) factory.Controller {
+
+	controllerFactory := factory.New()
+	target := &oauthAPIServerController{
+		targetNamespace:               targetNamespace,
+		oauthAPIServerTargetNamespace: oauthAPIServerTargetNamespace,
+		secretLister:                  kubeInformersForNamespaces.InformersFor(operatorclient.GlobalMachineSpecifiedConfigNamespace).Core().V1().Secrets().Lister().Secrets(operatorclient.GlobalMachineSpecifiedConfigNamespace),
+		secretClient:                  secretClient.Secrets(operatorclient.GlobalMachineSpecifiedConfigNamespace),
+	}
+
+	controllerFactory.WithSync(target.sync)
+	controllerFactory.WithInformers(kubeInformersForNamespaces.InformersFor(operatorclient.GlobalMachineSpecifiedConfigNamespace).Core().V1().Secrets().Informer())
+	return controllerFactory.ToController(name, eventRecorder.WithComponentSuffix("oauth-apiserver-encryption-cfg-sync-controller"))
+}
+
+// sync starts managing oauth-apiserver encryption config (encryption-config-openshift-oauth-apiserver in openshift-config-managed namespace) as described in https://github.com/openshift/enhancements/blob/master/enhancements/etcd/etcd-encryption-for-separate-oauth-apis.md
+//
+// case 1: if the secret doesn't exist and encryption is on then:
+//         - it will add the secret with the annotation (must be atomic operation) because CAO will start managing its own config iff the secret exist without the annotation
+//
+// case 2: if the secret exists and it is annotated
+//         - it will simply start synchronisation
+//
+// case 3: no-op: when the secret exits but it doesn't have the annotation - that means it was created by CAO in 4.6 and this is downgrade
+// case 4: no-op: when the secret doesn't exist and encryption is off
+//
+// drawbacks:
+// - it will not recover when the annotation was manually removed by a user,
+//   to recover we would have to put a value in the annotation instead and coordinate OAS-A and CAO but then we would have to remember to add it in CAO (4.6) as well
+func (c *oauthAPIServerController) sync(ctx context.Context, controllerContext factory.SyncContext) error {
+	openshiftAPIServerEncryptionCfg, err := c.secretLister.Get(fmt.Sprintf("%s-%s", encryptionconfig.EncryptionConfSecretName, c.targetNamespace))
+	if apierrors.IsNotFound(err) {
+		return nil // case 4: encryption off
+	}
+	if err != nil {
+		return err
+	}
+
+	oauthAPIServerEncryptionCfgName := fmt.Sprintf("%s-%s", encryptionconfig.EncryptionConfSecretName, c.oauthAPIServerTargetNamespace)
+	oauthAPIServerEncryptionCfg, err := c.secretLister.Get(oauthAPIServerEncryptionCfgName)
+	if err != nil && !apierrors.IsNotFound(err) {
+		return err
+	}
+	if apierrors.IsNotFound(err) {
+		// case 1: create with annotation
+		if _, exists := openshiftAPIServerEncryptionCfg.Data[encryptionconfig.EncryptionConfSecretKey]; !exists {
+			return fmt.Errorf("%s/%s doesn't contain the required key %q", openshiftAPIServerEncryptionCfg.Namespace, openshiftAPIServerEncryptionCfg.Name, encryptionconfig.EncryptionConfSecretKey)
+		}
+		encryptionCfg := &corev1.Secret{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      oauthAPIServerEncryptionCfgName,
+				Namespace: operatorclient.GlobalMachineSpecifiedConfigNamespace,
+				Annotations: map[string]string{
+					encryptionConfigManagedBy:                encryptionConfigManagedByValue,
+					encryptionstate.KubernetesDescriptionKey: encryptionstate.KubernetesDescriptionScaryValue,
+				},
+				Finalizers: []string{encryptionsecret.EncryptionSecretFinalizer},
+			},
+			Data: map[string][]byte{},
+		}
+		encryptionCfg.Data[encryptionconfig.EncryptionConfSecretKey] = openshiftAPIServerEncryptionCfg.Data[encryptionconfig.EncryptionConfSecretKey]
+
+		_, err := c.secretClient.Create(ctx, encryptionCfg, metav1.CreateOptions{})
+		if err != nil {
+			return err
+		}
+		controllerContext.Recorder().Eventf("SecretCreated", "Created %s in %s namespace because it was missing", oauthAPIServerEncryptionCfgName, operatorclient.GlobalMachineSpecifiedConfigNamespace)
+		return nil
+	}
+	if _, exist := oauthAPIServerEncryptionCfg.Annotations[encryptionConfigManagedBy]; exist {
+		// case 2: exists and it is annotated
+		oauthEncryptionCfgData := oauthAPIServerEncryptionCfg.Data[encryptionconfig.EncryptionConfSecretKey]
+		oasEncryptionCfgData := openshiftAPIServerEncryptionCfg.Data[encryptionconfig.EncryptionConfSecretKey]
+		if !equality.Semantic.DeepEqual(oauthEncryptionCfgData, oasEncryptionCfgData) {
+			encryptionCfg := oauthAPIServerEncryptionCfg.DeepCopy()
+			encryptionCfg.Data[encryptionconfig.EncryptionConfSecretName] = oasEncryptionCfgData
+			_, err := c.secretClient.Update(ctx, encryptionCfg, metav1.UpdateOptions{})
+			if err != nil {
+				return err
+			}
+			controllerContext.Recorder().Eventf("SecretUpdated", "Updates %s in %s namespace because it was out of date with %s ", oauthAPIServerEncryptionCfgName, operatorclient.GlobalMachineSpecifiedConfigNamespace, openshiftAPIServerEncryptionCfg.Name)
+			return nil
+		}
+		return nil
+	}
+
+	// case 3: no-op the secret is managed by CAO
+	return nil
+}
--- a/pkg/operator/oauthapiencryptioncontroller/oauthapiencryptioncontroller_test.go
+++ b/pkg/operator/oauthapiencryptioncontroller/oauthapiencryptioncontroller_test.go
@@ -1,0 +1,242 @@
+package oauthapiencryptioncontroller
+
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
+	"k8s.io/apimachinery/pkg/api/meta"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/diff"
+	"k8s.io/client-go/kubernetes/fake"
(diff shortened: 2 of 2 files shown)


pkg/operator/oauthapiencryptioncontroller/oauthapiencryptioncontroller.go
```diff
@@ -0,0 +1,128 @@
...
+
+	secretLister corev1listers.SecretNamespaceLister
+	secretClient corev1client.SecretsGetter
+
+	synchronizationEnabled bool
+	resourceSyncer         resourcesynccontroller.ResourceSyncer
+}
+
+// New creates OAuthAPIServerController that will manage encryption-config-openshift-oauth-apiserver as described in https://github.com/openshift/enhancements/blob/master/enhancements/etcd/etcd-encryption-for-separate-oauth-apis.md
+func New(
```
p0lyn0mial: do we want to add a `BZ` so that we don't forget to remove that code in `4.6` ?

sttts: yes please

pkg/operator/oauthapiencryptioncontroller/oauthapiencryptioncontroller.go
```diff
@@ -0,0 +1,128 @@
...
+	oauthAPIServerTargetNamespace string
+
+	secretLister corev1listers.SecretNamespaceLister
+	secretClient corev1client.SecretsGetter
+
+	synchronizationEnabled bool
+	resourceSyncer         resourcesynccontroller.ResourceSyncer
+}
+
+// New creates OAuthAPIServerController that will manage encryption-config-openshift-oauth-apiserver as described in https://github.com/openshift/enhancements/blob/master/enhancements/etcd/etcd-encryption-for-separate-oauth-apis.md
```
sttts: add namespace maybe

pkg/operator/oauthapiencryptioncontroller/oauthapiencryptioncontroller.go
```diff
@@ -0,0 +1,128 @@
...
+		secretClient:                  client,
+		resourceSyncer:                resourceSyncer,
+	}
+
+	controllerFactory.WithSync(target.sync)
+	controllerFactory.WithInformers(kubeInformersForNamespaces.InformersFor(operatorclient.GlobalMachineSpecifiedConfigNamespace).Core().V1().Secrets().Informer())
+	return controllerFactory.ToController(name, eventRecorder.WithComponentSuffix("oauth-apiserver-encryption-cfg-sync-controller"))
+}
+
+// sync start managing oauth-apiserver encryption config (encryption-config-openshift-oauth-apiserver in openshift-config-managed namespace) as described in https://github.com/openshift/enhancements/blob/master/enhancements/etcd/etcd-encryption-for-separate-oauth-apis.md
```
sttts: nit: start*s*

pkg/operator/oauthapiencryptioncontroller/oauthapiencryptioncontroller.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+
+	oauthAPIServerEncryptionCfg, err := c.secretLister.Get(fmt.Sprintf("%s-%s", encryptionconfig.EncryptionConfSecretName, c.oauthAPIServerTargetNamespace))
+	if err != nil && !apierrors.IsNotFound(err) {
+		return err
+	}
+	if apierrors.IsNotFound(err) {
+		// case 1: create with annotation and start synchronization
+		source := openshiftAPIServerEncryptionCfg.DeepCopy()
+		source.Name = fmt.Sprintf("%s-%s", encryptionconfig.EncryptionConfSecretName, c.oauthAPIServerTargetNamespace)
```
sttts: factor out to, used 10 lines above too

pkg/operator/oauthapiencryptioncontroller/oauthapiencryptioncontroller.go
```diff
@@ -0,0 +1,128 @@
...
+		source.ResourceVersion = ""
+		source.OwnerReferences = []metav1.OwnerReference{}
```
sttts: would rather start with a fresh ObjectMeta and copy everything over you want. There are other fields we probably don't want here.

sttts: finalizers, managedFields, customer annotations, labels, ...

pkg/operator/oauthapiencryptioncontroller/oauthapiencryptioncontroller.go
```diff
@@ -0,0 +1,128 @@
...
+		if source.Annotations == nil {
+			source.Annotations = map[string]string{}
+		}
+		source.Annotations[encryptionConfigManagedBy] = encryptionConfigManagedByValue
+		if _, _, err = resourceapply.ApplySecret(c.secretClient, controllerContext.Recorder(), source); err != nil {
```
sttts: isn't this apply racy? Would prefer to see more primitive use of the client here.

pkg/operator/oauthapiencryptioncontroller/oauthapiencryptioncontroller.go
```diff
@@ -0,0 +1,128 @@
...
+	if _, exist := oauthAPIServerEncryptionCfg.Annotations[encryptionConfigManagedBy]; exist {
+		// case 2: exists and it is annotated
+		return c.startSecretSynchronization(controllerContext.Recorder())
+	}
+
+	// case 3: no-op the secret is managed by CAO
+	return nil
+}
+
+func (c *oauthAPIServerController) startSecretSynchronization(recorder events.Recorder) error {
```
sttts: start suggests this is a background task which it isn't.

pkg/operator/oauthapiencryptioncontroller/oauthapiencryptioncontroller.go
```diff
@@ -0,0 +1,128 @@
...
+		return nil
+	}
+
+	if err := c.resourceSyncer.SyncSecret(
+		resourcesynccontroller.ResourceLocation{Namespace: operatorclient.GlobalMachineSpecifiedConfigNamespace, Name: fmt.Sprintf("%s-%s", encryptionconfig.EncryptionConfSecretName, c.oauthAPIServerTargetNamespace)},
+		resourcesynccontroller.ResourceLocation{Namespace: operatorclient.GlobalMachineSpecifiedConfigNamespace, Name: fmt.Sprintf("%s-%s", encryptionconfig.EncryptionConfSecretName, c.targetNamespace)},
+	); err != nil {
+		return err
+	}
+	c.synchronizationEnabled = true
```
sttts: why do we need this state? It's surprising.

pkg/operator/oauthapiencryptioncontroller/oauthapiencryptioncontroller.go
```diff
@@ -0,0 +1,128 @@
...
+}
+
+func (c *oauthAPIServerController) startSecretSynchronization(recorder events.Recorder) error {
+	if c.synchronizationEnabled {
+		return nil
+	}
+
+	if err := c.resourceSyncer.SyncSecret(
+		resourcesynccontroller.ResourceLocation{Namespace: operatorclient.GlobalMachineSpecifiedConfigNamespace, Name: fmt.Sprintf("%s-%s", encryptionconfig.EncryptionConfSecretName, c.oauthAPIServerTargetNamespace)},
+		resourcesynccontroller.ResourceLocation{Namespace: operatorclient.GlobalMachineSpecifiedConfigNamespace, Name: fmt.Sprintf("%s-%s", encryptionconfig.EncryptionConfSecretName, c.targetNamespace)},
```
sttts: also here, better use client primitives. It's easier to reason about than these with less clear semantics.

pkg/operator/oauthapiencryptioncontroller/oauthapiencryptioncontroller.go
p0lyn0mial: I need to `atomically` create a secret with the annotation because there will be an active process `CAO` waiting for it (It will take over If it doesn't have the annotation)

I think that `resourceapply.ApplySecret` guarantees atomicity.

pkg/operator/oauthapiencryptioncontroller/oauthapiencryptioncontroller.go
p0lyn0mial: `resourceSyncer.SyncSecret` will keep two secrets in `sync`

pkg/operator/oauthapiencryptioncontroller/oauthapiencryptioncontroller.go
p0lyn0mial: I don't want to call `c.resourceSyncer.SyncSecret` multiple times - only on startup or when encryption has been turned on.

pkg/operator/oauthapiencryptioncontroller/oauthapiencryptioncontroller.go
sttts: I tought it does clever merging, i.e. adding the annotation if the secret is pre-existing.

pkg/operator/oauthapiencryptioncontroller/oauthapiencryptioncontroller.go
sttts: this is a one-time call?

sttts: i.e. no start of a controller. Then maybe it's fine.

pkg/operator/oauthapiencryptioncontroller/oauthapiencryptioncontroller.go
sttts: I thought we have to call it continuously

pkg/operator/oauthapiencryptioncontroller/oauthapiencryptioncontroller.go
p0lyn0mial: yes, this is a one-time call, protected by `c.synchronizationEnabled`

pkg/operator/oauthapiencryptioncontroller/oauthapiencryptioncontroller.go
p0lyn0mial: since I'm relying on `c.resourceSyncer.SyncSecret` annotations and labels will be copied anyway but I get rid of finalizers.

pkg/operator/oauthapiencryptioncontroller/oauthapiencryptioncontroller.go
p0lyn0mial: `c.resourceSyncer` is a controller.

pkg/operator/oauthapiencryptioncontroller/oauthapiencryptioncontroller.go
p0lyn0mial: it will merge labels and annotations but someone needs to create the annotation first, this is what we do in this PR.

sttts: the other operator creates the object without annotation. You apply your version to it and the apply code adds the annotations. It shouldn't.

pkg/operator/oauthapiencryptioncontroller/oauthapiencryptioncontroller.go
sttts: I don't like state in controllers which is changed during the controller runtime. It's a bad pattern leading to bugs IMO.

pkg/operator/oauthapiencryptioncontroller/oauthapiencryptioncontroller.go
sttts: How do we stop it again?

pkg/operator/oauthapiencryptioncontroller/oauthapiencryptioncontroller.go
p0lyn0mial: which operator? 
In `4.5` only this operator `OAS-O` will be creating and managing the secret.
In `4.6` only `CAO` will be creating and managing the secret.

pkg/operator/oauthapiencryptioncontroller/oauthapiencryptioncontroller.go
p0lyn0mial: we don't, In `4.6` we will remove the code, on downgrades (the secret exist but doesn't have the annotation) we do nothing.

pkg/operator/oauthapiencryptioncontroller/oauthapiencryptioncontroller.go
sttts: and with skew? I don't like this non-obvious assumptions. Maybe they are right, but they are not explicit here. A simple create call with a fresh ObjectMeta and it's trivial to reason about.

p0lyn0mial: isn't it what we wanted?

`Step 1` from https://github.com/openshift/enhancements/blob/master/enhancements/etcd/etcd-encryption-for-separate-oauth-apis.md:

```
Create a new controller in OAS-O that will create and annotate encryption-config-openshift-oauth secret in openshift-config-managed namespace.
It also must keep it in sync with encryption-config-openshift-apiserver. 
The annotation will prevent CAO from managing the newly created encryption-config
```

Not sure if I understand what you are proposing.

pkg/operator/oauthapiencryptioncontroller/oauthapiencryptioncontroller.go
p0lyn0mial: done https://bugzilla.redhat.com/show_bug.cgi?id=1817629

pkg/operator/oauthapiencryptioncontroller/oauthapiencryptioncontroller.go
```diff
@@ -0,0 +1,129 @@
...
+
+	oauthAPIServerEncryptionCfgName := fmt.Sprintf("%s-%s", encryptionconfig.EncryptionConfSecretName, c.oauthAPIServerTargetNamespace)
+	oauthAPIServerEncryptionCfg, err := c.secretLister.Get(oauthAPIServerEncryptionCfgName)
+	if err != nil && !apierrors.IsNotFound(err) {
+		return err
+	}
+	if apierrors.IsNotFound(err) {
+		// case 1: create with annotation
+		if _, exists := openshiftAPIServerEncryptionCfg.Data[encryptionconfig.EncryptionConfSecretKey]; !exists {
+			return fmt.Errorf("%s/%s doesn't contain the required key %q", openshiftAPIServerEncryptionCfg.Name, openshiftAPIServerEncryptionCfg.Namespace, encryptionconfig.EncryptionConfSecretKey)
```
sttts: swap ns and name

pkg/operator/oauthapiencryptioncontroller/oauthapiencryptioncontroller.go
```diff
@@ -0,0 +1,129 @@
...
+		}
+		encryptionCfg := &corev1.Secret{}
+		encryptionCfg.Name = oauthAPIServerEncryptionCfgName
+		encryptionCfg.Namespace = operatorclient.GlobalMachineSpecifiedConfigNamespace
+		encryptionCfg.Annotations = map[string]string{}
+		encryptionCfg.Annotations[encryptionConfigManagedBy] = encryptionConfigManagedByValue
+		encryptionCfg.Annotations[encryptionstate.KubernetesDescriptionKey] = encryptionstate.KubernetesDescriptionScaryValue
+		encryptionCfg.Data = map[string][]byte{}
+		encryptionCfg.Data[encryptionconfig.EncryptionConfSecretKey] = openshiftAPIServerEncryptionCfg.Data[encryptionconfig.EncryptionConfSecretKey]
+		encryptionCfg.Finalizers = []string{encryptionsecret.EncryptionSecretFinalizer}
```
sttts:
```golang
encryptionCfg := &corev1.Secret{
  ObjectMeta: metav1.ObjectMeta{
    Name: ...
    Namespcae: ...
  },
  ...
}
```

is easier to read IMO

sttts: is there a controller doing finalizing? Or is it just there to block deletion?

pkg/operator/oauthapiencryptioncontroller/oauthapiencryptioncontroller.go
```diff
@@ -0,0 +1,129 @@
...
+			return fmt.Errorf("%s/%s doesn't contain the required key %q", openshiftAPIServerEncryptionCfg.Name, openshiftAPIServerEncryptionCfg.Namespace, encryptionconfig.EncryptionConfSecretKey)
+		}
+		encryptionCfg := &corev1.Secret{}
+		encryptionCfg.Name = oauthAPIServerEncryptionCfgName
+		encryptionCfg.Namespace = operatorclient.GlobalMachineSpecifiedConfigNamespace
+		encryptionCfg.Annotations = map[string]string{}
+		encryptionCfg.Annotations[encryptionConfigManagedBy] = encryptionConfigManagedByValue
+		encryptionCfg.Annotations[encryptionstate.KubernetesDescriptionKey] = encryptionstate.KubernetesDescriptionScaryValue
+		encryptionCfg.Data = map[string][]byte{}
+		encryptionCfg.Data[encryptionconfig.EncryptionConfSecretKey] = openshiftAPIServerEncryptionCfg.Data[encryptionconfig.EncryptionConfSecretKey]
```
sttts: nit: why not whole Data?

pkg/operator/oauthapiencryptioncontroller/oauthapiencryptioncontroller.go
```diff
@@ -0,0 +1,129 @@
...
+		controllerContext.Recorder().Eventf("SecretCreated", "Created %s in %s namespace because it was missing", oauthAPIServerEncryptionCfgName, operatorclient.GlobalMachineSpecifiedConfigNamespace)
+		return nil
+	}
+	if _, exist := oauthAPIServerEncryptionCfg.Annotations[encryptionConfigManagedBy]; exist {
+		// case 2: exists and it is annotated
+		oauthEncryptionCfgData := oauthAPIServerEncryptionCfg.Data[encryptionconfig.EncryptionConfSecretKey]
+		oasEncryptionCfgData := openshiftAPIServerEncryptionCfg.Data[encryptionconfig.EncryptionConfSecretKey]
+		if !equality.Semantic.DeepEqual(oauthEncryptionCfgData, oasEncryptionCfgData) {
+			encryptionCfg := oauthAPIServerEncryptionCfg.DeepCopy()
+			encryptionCfg.Data[encryptionconfig.EncryptionConfSecretKey] = oasEncryptionCfgData
```
sttts: didn't we have any annotations that play a role? Or just for keys?

pkg/operator/oauthapiencryptioncontroller/oauthapiencryptioncontroller.go
p0lyn0mial: I added it to block deletion.

pkg/operator/oauthapiencryptioncontroller/oauthapiencryptioncontroller.go
p0lyn0mial: it's a map and someone else could have added entries we don't need.

pkg/operator/oauthapiencryptioncontroller/oauthapiencryptioncontroller.go
p0lyn0mial: we actually use https://github.com/openshift/library-go/blob/master/pkg/operator/encryption/encryptionconfig/secret.go#L71 but the content is the same :) (uff)

PR #347, opened 2020-03-31
encryption provider

p0lyn0mial:


pkg/operator/encryptionprovider/encryptionprovider.go
```diff
@@ -0,0 +1,93 @@
...
+		secretLister:                        kubeInformersForNamespaces.InformersFor(operatorclient.GlobalMachineSpecifiedConfigNamespace).Core().V1().Secrets().Lister().Secrets(operatorclient.GlobalMachineSpecifiedConfigNamespace),
+		eventRecorder:                       eventRecorder,
+	}
+}
+
+// EncryptedGRs returns resources that need to be encrypted
+func (p *encryptionProvider) EncryptedGRs() []schema.GroupResource {
+	oauthAPIServerEncryptionCfg, err := p.secretLister.Get(fmt.Sprintf("%s-%s", encryptionconfig.EncryptionConfSecretName, p.oauthAPIServerTargetNamespace))
+	if err != nil {
+		return p.defaultEncryptedGRs
```
sttts, line 53: I don't like this. If it is a NotFound, I agree with this. Can we be sure there are no other errors? Just returning the default list feels unsafe.

sttts: Also note that before encryption is enabled, there is no secret. We would return an incomplete list in 4.5 here.

sttts: oh wait, the default list is all? This is a bad name.

pkg/operator/encryptionprovider/encryptionprovider.go
```diff
@@ -0,0 +1,93 @@
...
+	newEncryptedGRsToManage := []schema.GroupResource{}
+	for _, gr := range p.defaultEncryptedGRs {
+		if p.encryptedGRsManagedByExternalServer.Has(gr.String()) {
+			continue
+		}
+		newEncryptedGRsToManage = append(newEncryptedGRsToManage, gr)
+	}
+	if changed, newEncryptedGRsToManageSet := haveGRsChanged(p.currentEncryptedGRs, newEncryptedGRsToManage); changed {
+		p.eventRecorder.Eventf("EncryptedGRsChanged", "The new GroupResource list this operator will manage is %v", newEncryptedGRsToManageSet.List())
+		p.currentEncryptedGRs = newEncryptedGRsToManage
```
sttts, line 69: this is wrong spot. Also state like this is an anti-pattern. Side-effects from a getter like this are very unexpected.

pkg/operator/encryptionprovider/encryptionprovider.go
p0lyn0mial: what do you want to return on errors?
when can lister return an error?

returning the previously observed state can cause flapping

pkg/operator/encryptionprovider/encryptionprovider.go
p0lyn0mial: which getter? the value is read from the cache and won't change once set.

pkg/operator/encryptionprovider/encryptionprovider.go
p0lyn0mial: that's why I would prefer to set it once on startup - now the value can change during runtime and it shouldn't.

pkg/operator/encryptionprovider/encryptionprovider.go
p0lyn0mial: the state (`currentEncryptedGRs`) was introduced to emit an event only once it has changed.

pkg/operator/encryptionprovider/encryptionprovider.go
p0lyn0mial: > oh wait, the default list is all? This is a bad name.

oh man, It's the default list this operator starts with :)
would you like me to rename to `allEncryptedGRs` ?

pkg/operator/encryptionprovider/encryptionprovider.go
p0lyn0mial: > this is wrong spot.

why ? we assign to `currentEncryptedGRs` only when it has changed.

sttts: Wasn't clear with my comment. Emitting events from a getter like this seems to be wrong.

p0lyn0mial: wouldn't it be nice to know that the list have changed (debuggability)? 
frequent events might indicate errors.

pkg/operator/encryptionprovider/encryptionprovider.go
```diff
@@ -67,37 +61,18 @@ func (p *encryptionProvider) EncryptedGRs() []schema.GroupResource {
 		return p.allEncryptedGRs // case 1 - we are in charge
 	}
 
+	// case 2 - CAO is in charge, reduce the list
```
p0lyn0mial, line 64: we could make the reduced list to be static too

sttts: or have a subtract func

PR #350, opened 2020-04-13
Bug 1821771: Remove unauthenticated user group from discover cluster role bindings

vareti:
When a cluster is upgraded from 4.1 - > 4.4, `/` endpoint will be accessible for unauthenticated users. In 4.1 cluster, `/` resource endpoint is used in cluster roles which were given access to unauthenticated user group using `system:openshift:discovery` and `cluster-status-binding` cluster role bindings. After upgrade, as unauthenticated users are still part of those cluster role bindings and their access to the endpoint is not revoked. This PR aims to remove this group of users from cluster role bindings.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -20,6 +20,7 @@ require (
 	go.uber.org/multierr v1.1.1-0.20180122172545-ddea229ff1df // indirect
 	k8s.io/api v0.18.2
 	k8s.io/apimachinery v0.18.2
+	k8s.io/apiserver v0.18.2
 	k8s.io/client-go v0.18.2
 	k8s.io/component-base v0.18.2
 	k8s.io/klog v1.0.0
@@ -27,4 +28,7 @@ require (
 	k8s.io/utils v0.0.0-20200324210504-a9aa75ae1b89
 )
 
-replace github.com/kubernetes-sigs/kube-storage-version-migrator => github.com/openshift/kubernetes-kube-storage-version-migrator v0.0.3-0.20200312103335-32e07ea4f8ca
+replace (
+	github.com/kubernetes-sigs/kube-storage-version-migrator => github.com/openshift/kubernetes-kube-storage-version-migrator v0.0.3-0.20200312103335-32e07ea4f8ca
+	github.com/openshift/library-go => github.com/mfojtik/library-go v0.0.0-20200511161500-a4b1c3e3a7a6
+)
--- a/pkg/operator/rbaccontroller/rbaccontroller.go
+++ b/pkg/operator/rbaccontroller/rbaccontroller.go
@@ -1,0 +1,126 @@
+package rbaccontroller
+
+import (
+	"context"
+
+	"github.com/openshift/library-go/pkg/controller/factory"
+	"github.com/openshift/library-go/pkg/operator/events"
+
+	rbacv1 "k8s.io/api/rbac/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/apiserver/pkg/authentication/user"
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/kubernetes"
+	listerrbacv1 "k8s.io/client-go/listers/rbac/v1"
+	"k8s.io/client-go/tools/cache"
+)
+
+const (
+	// This annotation is set to a value of "true" on RBAC resources created by the openshift-apiserver.
+	openShiftManagedRBACAnnotation = "openshiftapiserver.rbac.authorization.openshift.io/managed"
+)
+
+var (
+	// The named cluster role bindings are created by openshift-apiserver
+	// and in previous releases may have included the unauthenticated subject.
+	clusterRoleBindingNames = sets.NewString("discovery", "basic-users", "system:openshift:discovery",
+		"cluster-status-binding")
+)
+
+// RBACController reconciles RBAC resources created by openshift-apiserver.
+//
+// If openshiftManagedRBACAnnotation is set to "true" for a cluster role binding,
+// RBACController ensures the removal of the unauthenticated subject from its list of subjects.
+//
+// A user can opt-out of this reconciliation by setting the annotation to "false".
+type RBACController struct {
+	eventRecorder            events.Recorder
+	kubeClient               kubernetes.Interface
+	clusterRoleBindingLister listerrbacv1.ClusterRoleBindingLister
+}
+
+func NewRBACController(kubeClient kubernetes.Interface, informerFactory informers.SharedInformerFactory, recorder events.Recorder) factory.Controller {
+	c := &RBACController{
+		kubeClient:               kubeClient,
+		clusterRoleBindingLister: informerFactory.Rbac().V1().ClusterRoleBindings().Lister(),
+		eventRecorder:            recorder,
+	}
+
+	syncCtx := factory.NewSyncContext("RBACController", c.eventRecorder)
+
+	informer := informerFactory.Rbac().V1().ClusterRoleBindings().Informer()
+	informer.AddEventHandler(cache.FilteringResourceEventHandler{
+		FilterFunc: hasMatchingAttributes,
+		Handler: cache.ResourceEventHandlerFuncs{
+			AddFunc: func(obj interface{}) {
+				key, err := cache.MetaNamespaceKeyFunc(obj)
+				if err == nil {
+					syncCtx.Queue().Add(key)
+				}
+			},
+			UpdateFunc: func(oldObj, newObj interface{}) {
+				key, err := cache.MetaNamespaceKeyFunc(newObj)
+				if err == nil {
+					syncCtx.Queue().Add(key)
+				}
+			},
+		},
+	})
+
+	return factory.New().
+		WithBareInformers(informer).
+		WithSyncContext(syncCtx).
+		WithSync(c.sync).
+		ToController("", nil)
+}
+
+// sync ensures that cluster role bindings created by openshift-apiserver do not have the unauthenticated subject.
+func (c *RBACController) sync(ctx context.Context, syncContext factory.SyncContext) error {
+	_, clusterRoleBindingName, err := cache.SplitMetaNamespaceKey(syncContext.QueueKey())
+	if err != nil {
+		return err
+	}
+
+	clusterRoleBinding, err := c.clusterRoleBindingLister.Get(clusterRoleBindingName)
+	if err != nil {
+		return err
+	}
+
+	updatedCopy := withUnauthenticatedSubjectRemoved(clusterRoleBinding)
+	if updatedCopy == nil {
+		return nil
+	}
+
+	_, err = c.kubeClient.RbacV1().ClusterRoleBindings().Update(ctx, updatedCopy, metav1.UpdateOptions{})
+	return err
+}
+
+// withUnauthenticatedSubjectRemoved returns a new copy of given cluster role binding without unauthenticated user in subjects.
+func withUnauthenticatedSubjectRemoved(clusterRoleBinding *rbacv1.ClusterRoleBinding) *rbacv1.ClusterRoleBinding {
+	for i, subject := range clusterRoleBinding.Subjects {
+		if subject.Name == user.AllUnauthenticated {
+			clusterRoleBindingCopy := clusterRoleBinding.DeepCopy()
+			clusterRoleBindingCopy.Subjects = append(clusterRoleBindingCopy.Subjects[:i],
+				clusterRoleBindingCopy.Subjects[i+1:]...)
+			return clusterRoleBindingCopy
+		}
+	}
+
+	return nil
+}
+
+// hasMatchingAttributes returns true if the obj name equals one of
+// clusterRoleBindingNames and it's RBAC annotation is set to "true".
+func hasMatchingAttributes(obj interface{}) bool {
+	clusterRoleBinding, ok := obj.(*rbacv1.ClusterRoleBinding)
+	if !ok {
+		return false
+	}
+
+	if !clusterRoleBindingNames.Has(clusterRoleBinding.Name) {
+		return false
+	}
+
+	return (clusterRoleBinding.Annotations[openShiftManagedRBACAnnotation] == "true")
+}
--- a/pkg/operator/starter.go
+++ b/pkg/operator/starter.go
@@ -13,6 +13,7 @@ import (
 	configinformers "github.com/openshift/client-go/config/informers/externalversions"
 	operatorv1client "github.com/openshift/client-go/operator/clientset/versioned"
 	operatorv1informers "github.com/openshift/client-go/operator/informers/externalversions"
+	"github.com/openshift/cluster-openshift-apiserver-operator/pkg/operator/rbaccontroller"
 	"github.com/openshift/library-go/pkg/controller/controllercmd"
 	"github.com/openshift/library-go/pkg/operator/encryption/controllers/migrators"
 	encryptiondeployer "github.com/openshift/library-go/pkg/operator/encryption/deployer"
@@ -251,6 +252,10 @@ func RunOperator(ctx context.Context, controllerConfig *controllercmd.Controller
 		controllerConfig.EventRecorder,
 	)
 
+	rbacController := rbaccontroller.NewRBACController(kubeClient,
+		kubeInformersForNamespaces.InformersFor(""),
+		controllerConfig.EventRecorder)
+
 	staleConditions := staleconditions.NewRemoveStaleConditionsController(
 		[]string{
 			// in 4.1.0-4.3.0 this was used for indicating the apiserver daemonset was progressing
@@ -287,6 +292,7 @@ func RunOperator(ctx context.Context, controllerConfig *controllercmd.Controller
 	go pruneController.Run(ctx, 1)
 	go runnableAPIServerControllers.Run(ctx)
 	go staleConditions.Run(ctx, 1)
+	go rbacController.Run(ctx, 1)
 
 	<-ctx.Done()
 	return nil
--- a/go.sum
+++ b/go.sum
@@ -279,6 +279,8 @@ github.com/mattn/go-isatty v0.0.4/go.mod h1:M+lRXTBqGeGNdLjl/ufCoiOlB5xdOkqRJdNx
 github.com/mattn/go-runewidth v0.0.2/go.mod h1:LwmH8dsx7+W8Uxz3IHJYH5QSwggIsqBzpuz5H//U1FU=
 github.com/matttproud/golang_protobuf_extensions v1.0.1 h1:4hp9jkHxhMHkqkrB3Ix0jegS5sx/RkqARlsWZ6pIwiU=
 github.com/matttproud/golang_protobuf_extensions v1.0.1/go.mod h1:D8He9yQNgCq6Z5Ld7szi9bcBfOoFv/3dc6xSMkL2PC0=
+github.com/mfojtik/library-go v0.0.0-20200511161500-a4b1c3e3a7a6 h1:RDl00jaxal+AHtFOKuEeofz5DjIK+DslhvvA7Ixed+A=
+github.com/mfojtik/library-go v0.0.0-20200511161500-a4b1c3e3a7a6/go.mod h1:2kWwXTkpoQJUN3jZ3QW88EIY1hdRMqxgRs2hheEW/pg=
 github.com/mitchellh/go-homedir v1.1.0/go.mod h1:SfyaCUpYCn1Vlf4IUYiD9fPX4A5wJrkLzIz1N1q0pr0=
 github.com/mitchellh/mapstructure v1.1.2 h1:fmNYVwqnSfB9mZU6OS2O6GsXM+wcskZDuKQzvN1EDeE=
 github.com/mitchellh/mapstructure v1.1.2/go.mod h1:FVVH3fgwuzCH5S8UJGiWEs2h04kUh9fWfEaFds41c1Y=
@@ -318,8 +320,6 @@ github.com/openshift/client-go v0.0.0-20200326155132-2a6cd50aedd0 h1:kMiuiZXH1Gd
 github.com/openshift/client-go v0.0.0-20200326155132-2a6cd50aedd0/go.mod h1:uUQ4LClRO+fg5MF/P6QxjMCb1C9f7Oh4RKepftDnEJE=
 github.com/openshift/kubernetes-kube-storage-version-migrator v0.0.3-0.20200312103335-32e07ea4f8ca h1:YNtyJnE53QuEUSjl7L1AARocI021o7cU2bvh4prDtiE=
 github.com/openshift/kubernetes-kube-storage-version-migrator v0.0.3-0.20200312103335-32e07ea4f8ca/go.mod h1:unEnEWccGeVxaXSRsWTjRsNxMqYXmuQjzjcPFQ91H9M=
-github.com/openshift/library-go v0.0.0-20200427130628-9b02543ac833 h1:glEuGTvwkQJXfCzVfosZ3oq73L8QltO6Xyxo1oTf5Ig=
-github.com/openshift/library-go v0.0.0-20200427130628-9b02543ac833/go.mod h1:2kWwXTkpoQJUN3jZ3QW88EIY1hdRMqxgRs2hheEW/pg=
 github.com/pborman/uuid v1.2.0 h1:J7Q5mO4ysT1dv8hyrUGHb9+ooztCXu1D8MY8DZYsu3g=
 github.com/pborman/uuid v1.2.0/go.mod h1:X/NO0urCmaxf9VXbdlT7C2Yzkj2IKimNn4k+gtPdI/k=
 github.com/pelletier/go-toml v1.2.0/go.mod h1:5z9KED0ma1S8pY6P1sdut58dfprrGBbd/94hg7ilaic=
--- a/pkg/operator/rbaccontroller/rbaccontroller_test.go
+++ b/pkg/operator/rbaccontroller/rbaccontroller_test.go
@@ -1,0 +1,116 @@
+package rbaccontroller
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/openshift/library-go/pkg/operator/events"
+
+	rbacv1 "k8s.io/api/rbac/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apiserver/pkg/authentication/user"
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/kubernetes/fake"
+)
+
+func TestReconcileClusterRoleBinding(t *testing.T) {
+
+	tests := []struct {
+		clusterRoleBinding rbacv1.ClusterRoleBinding
+		expNumOfSubjects   int
+	}{
+		{
+			clusterRoleBinding: rbacv1.ClusterRoleBinding{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:        "basic-users",
+					Annotations: map[string]string{openShiftManagedRBACAnnotation: "true"},
+				},
+				Subjects: []rbacv1.Subject{
+					{
+						Name: user.AllUnauthenticated,
+					},
+					{
+						Name: user.AllAuthenticated,
+					},
+				},
+			},
+			expNumOfSubjects: 1,
+		},
+		{
+			clusterRoleBinding: rbacv1.ClusterRoleBinding{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:        "discovery",
+					Annotations: map[string]string{openShiftManagedRBACAnnotation: "true"},
+				},
+				Subjects: []rbacv1.Subject{
+					{
+						Name: user.AllAuthenticated,
+					},
+				},
+			},
+			expNumOfSubjects: 1,
+		},
+		{
+			clusterRoleBinding: rbacv1.ClusterRoleBinding{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:        "system:openshift:discovery",
+					Annotations: map[string]string{openShiftManagedRBACAnnotation: "false"},
+				},
+				Subjects: []rbacv1.Subject{
+					{
+						Name: user.AllUnauthenticated,
+					},
+				},
+			},
+			expNumOfSubjects: 1,
+		},
+		{
+			clusterRoleBinding: rbacv1.ClusterRoleBinding{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "cluster-status-binding",
+				},
+				Subjects: []rbacv1.Subject{
+					{
+						Name: user.AllUnauthenticated,
+					},
+				},
+			},
+			expNumOfSubjects: 1,
+		},
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	fakeClient := fake.NewSimpleClientset()
+	for _, obj := range tests {
+		_, err := fakeClient.RbacV1().ClusterRoleBindings().Create(ctx, &obj.clusterRoleBinding, metav1.CreateOptions{})
+		if err != nil {
+			t.Fatal(err)
+		}
+	}
+
+	informersFactory := informers.NewSharedInformerFactory(fakeClient, 10*time.Minute)
+
+	recorder := events.NewInMemoryRecorder("rbac_controller_test")
+	c := NewRBACController(fakeClient, informersFactory, recorder)
+
+	informersFactory.Start(ctx.Done())
+
+	go c.Run(ctx, 1)
+
+	time.Sleep(1 * time.Second) // allowing for reconcile to happen
+
+	for _, obj := range tests {
+		clusterRoleBinding, err := fakeClient.RbacV1().ClusterRoleBindings().Get(ctx, obj.clusterRoleBinding.Name, metav1.GetOptions{})
+		if err != nil {
+			t.Fatal(err)
+		}
+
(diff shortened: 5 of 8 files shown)


pkg/operator/rbaccontroller/rbaccontroller.go
```diff
@@ -0,0 +1,78 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/apiserver/pkg/authentication/user"
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/klog"
+)
+
+type RBACController struct {
+	ctx           context.Context
```
vareti, line 19: remove this

pkg/operator/rbaccontroller/rbaccontroller.go
```diff
@@ -0,0 +1,78 @@
...
+	for _, clusterRoleBinding := range clusterRoleBindings {
+		if err := c.reconcileClusterRoleBinding(clusterRoleBinding); err != nil {
+			errs = append(errs, err)
+		}
+	}
+
+	return utilerrors.NewAggregate(errs)
+}
+
+func (c *RBACController) reconcileClusterRoleBinding(name string) error {
```
deads2k, line 57: This could create a new problem for us.   You need to have a way to opt-out of having the clusterrolebinding tightened.  We use an annotation in openshift/kube for "don't reconcile"

vareti: I see what you are saying. I will add a new annotation for these clusterrolebindings

vareti:
ci flakes

/retest

go.mod
```diff
@@ -18,15 +18,14 @@ require (
...
-	gonum.org/v1/netlib v0.0.0-20190331212654-76723241ea4e // indirect
 	k8s.io/api v0.18.0
 	k8s.io/apimachinery v0.18.0
+	k8s.io/apiserver v0.18.0
 	k8s.io/client-go v0.18.0
 	k8s.io/component-base v0.18.0
 	k8s.io/klog v1.0.0
 	k8s.io/kube-aggregator v0.18.0
 	k8s.io/utils v0.0.0-20200324210504-a9aa75ae1b89
-	sigs.k8s.io/structured-merge-diff v1.0.1-0.20191108220359-b1b620dd3f06 // indirect
```
sttts, line 29: split vendor updates from other changes

pkg/operator/rbaccontroller/rbaccontroller.go
```diff
@@ -0,0 +1,90 @@
...
+const (
+	// Bug 1821771: We needed a way to tighten the cluster role binding when cluster
+	// is upgraded from 4.1
+
+	// openShiftManagedRoleBindingAnnotation is a cluster role binding annotation used
+	// by operator to manage it when set to true
+	openShiftManagedRoleBindingAnnotation = "openshiftapiserver.rbac.authorization.openshift.io/managed"
+)
+
+type RBACController struct {
```
sttts, line 27: document what this does

pkg/operator/rbaccontroller/rbaccontroller.go
```diff
@@ -0,0 +1,90 @@
...
+	rbacv1 "k8s.io/api/rbac/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	errors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/apiserver/pkg/authentication/user"
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/kubernetes"
+)
+
+const (
+	// Bug 1821771: We needed a way to tighten the cluster role binding when cluster
```
sttts, line 19: we usually don't link BZs in code. PR description is enough.

pkg/operator/rbaccontroller/rbaccontroller.go
```diff
@@ -0,0 +1,90 @@
...
+	// is upgraded from 4.1
+
+	// openShiftManagedRoleBindingAnnotation is a cluster role binding annotation used
+	// by operator to manage it when set to true
```
sttts, line 23: used how?

pkg/operator/rbaccontroller/rbaccontroller.go
```diff
@@ -0,0 +1,90 @@
...
+		ToController("RBACController", c.eventRecorder)
+}
+
+func (c *RBACController) sync(ctx context.Context, syncContext factory.SyncContext) error {
+	clusterRoleBindingNames := []string{"discovery", "system:basic-user", "system:discovery", "system:openshift:discovery", "cluster-status-binding"}
+
+	var errs []error
+
+	for _, clusterRoleBindingName := range clusterRoleBindingNames {
+		clusterRoleBinding, err := c.kubeClient.RbacV1().ClusterRoleBindings().Get(context.Background(), clusterRoleBindingName, metav1.GetOptions{})
```
sttts, line 57: why background?

pkg/operator/rbaccontroller/rbaccontroller.go
```diff
@@ -0,0 +1,90 @@
...
+	return errors.NewAggregate(errs)
+}
+
+func reconcileClusterRoleBinding(clusterRoleBinding *rbacv1.ClusterRoleBinding) {
+	if clusterRoleBinding == nil {
+		return
+	}
+
+	val, ok := clusterRoleBinding.Annotations[openShiftManagedRoleBindingAnnotation]
+	if !ok || strings.Compare(val, "true") != 0 {
```
sttts, line 80: why `strings.Compare`?

pkg/operator/rbaccontroller/rbaccontroller.go
```diff
@@ -0,0 +1,90 @@
...
+		_, err = c.kubeClient.RbacV1().ClusterRoleBindings().Update(context.Background(), clusterRoleBinding, metav1.UpdateOptions{})
+		if err != nil {
+			errs = append(errs, err)
+		}
+	}
+
+	return errors.NewAggregate(errs)
+}
+
+func reconcileClusterRoleBinding(clusterRoleBinding *rbacv1.ClusterRoleBinding) {
```
sttts, line 74: what does it do?

pkg/operator/rbaccontroller/rbaccontroller.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+
+	val, ok := clusterRoleBinding.Annotations[openShiftManagedRoleBindingAnnotation]
+	if !ok || strings.Compare(val, "true") != 0 {
+		return
+	}
+
+	for i, subject := range clusterRoleBinding.Subjects {
+		if subject.Name == user.AllUnauthenticated {
+			clusterRoleBinding.Subjects = append(clusterRoleBinding.Subjects[:i], clusterRoleBinding.Subjects[i+1:]...)
```
sttts, line 86: changing the input object here is suspicious

pkg/operator/rbaccontroller/rbaccontroller.go
sttts, line 57: any reason we get these from the client and not the informer?

pkg/operator/starter.go
```diff
@@ -282,12 +289,14 @@ func RunOperator(ctx context.Context, controllerConfig *controllercmd.Controller
 	dynamicInformers.Start(ctx.Done())
 	migrationInformer.Start(ctx.Done())
 	migrationInformer.Start(ctx.Done())
```
sttts, line 291: unrelated (but pleaes fix in a separate PR: why migrationInformer twice?

pkg/operator/starter.go
```diff
@@ -251,6 +253,11 @@ func RunOperator(ctx context.Context, controllerConfig *controllercmd.Controller
 		controllerConfig.EventRecorder,
 	)
 
+	kubeInformers := informers.NewSharedInformerFactory(kubeClient, 10*time.Minute)
+	rbacController := rbaccontroller.NewRBACController(kubeClient,
+		kubeInformers,
+		controllerConfig.EventRecorder)
+
```
sttts, line 260: does kubeInformersForNamespaces work for cluster-wide resources too?

vareti: Namespace informer will not work for this case. As we are not limiting the informer to a namespace here, shouldn't it work in this case? Please correct me if I am wrong.

pkg/operator/starter.go
vareti: Yes. raised a PR https://github.com/openshift/cluster-openshift-apiserver-operator/pull/355/files

pkg/operator/rbaccontroller/rbaccontroller.go
vareti: Done! please check and let me know if it needs changes

go.mod
vareti: fixed now

pkg/operator/rbaccontroller/rbaccontroller.go
vareti: Done. Removed the link to BZ

pkg/operator/rbaccontroller/rbaccontroller.go
vareti: updated the description. please check

pkg/operator/rbaccontroller/rbaccontroller.go
vareti: it is a mistake on my part. fixed it now!

pkg/operator/rbaccontroller/rbaccontroller.go
vareti: was overthinking previously. switched to simple `==`

pkg/operator/rbaccontroller/rbaccontroller.go
vareti: added few sentences explaining what it does

pkg/operator/rbaccontroller/rbaccontroller.go
vareti: update the code to create a copy and then mutate it

pkg/operator/rbaccontroller/rbaccontroller.go
vareti: honestly, till date i thought both are same. read about the differences and update the code. thanks for pointing!

vareti:
@sttts update the PR addressing review comments. Please take a look whenever you are free

pkg/operator/rbaccontroller/rbaccontroller.go
```diff
@@ -0,0 +1,102 @@
...
+	"k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/apiserver/pkg/authentication/user"
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/kubernetes"
+	listerrbacv1 "k8s.io/client-go/listers/rbac/v1"
+	"k8s.io/klog"
+)
+
+const (
+	// On an upgrade from 4.1, some discovery cluster role bindings would
```
sttts, line 20: this description sounds very specific to the issue we had. That's good as background. But we need a crisp definition of what it does in general.

pkg/operator/rbaccontroller/rbaccontroller.go
sttts: "RBACController this controller" -> "RBACController removes..."

But same applies as for the annotation comment: it must be generic

pkg/operator/rbaccontroller/rbaccontroller.go
```diff
@@ -0,0 +1,102 @@
...
+	// to true, this rbac controller would remove unauthenticated user
+	// group from list of subjects from these cluster role bindings
+	openShiftManagedRoleBindingAnnotation = "openshiftapiserver.rbac.authorization.openshift.io/managed"
+)
+
+// RBACController this controller would remove unauthenticated user
+// group from list of subjects from some discovery cluster role bindings
+type RBACController struct {
+	kubeClient    kubernetes.Interface
+	rbacInformer  listerrbacv1.ClusterRoleBindingLister
```
sttts, line 32: better call it lister

pkg/operator/rbaccontroller/rbaccontroller.go
```diff
@@ -0,0 +1,102 @@
...
+	if !ok || val != "true" {
+		return clusterRoleBinding, false
+	}
+
+	clusterRoleBindingCopy := clusterRoleBinding.DeepCopy()
+
+	for i, subject := range clusterRoleBindingCopy.Subjects {
+		if subject.Name == user.AllUnauthenticated {
+			clusterRoleBindingCopy.Subjects = append(clusterRoleBindingCopy.Subjects[:i], clusterRoleBindingCopy.Subjects[i+1:]...)
+			break
```
sttts, line 97: why not return here with true? And with false below.

pkg/operator/rbaccontroller/rbaccontroller.go
```diff
@@ -0,0 +1,102 @@
...
+		}
+	}
+
+	return errors.NewAggregate(errs)
+}
+
+// reconcileClusterRoleBinding returns a new copy of cluster role binding after removing
+// unauthenticated user group, if present, from the list of subjects if
+// openShiftManagedRoleBindingAnnotation is set to true in the object passed
+func reconcileClusterRoleBinding(clusterRoleBinding *rbacv1.ClusterRoleBinding) (*rbacv1.ClusterRoleBinding, bool) {
```
sttts, line 82: would call this `withoutUnauthenticatedSubject`

pkg/operator/rbaccontroller/rbaccontroller.go
```diff
@@ -0,0 +1,102 @@
...
+	if clusterRoleBinding == nil {
+		return nil, false
+	}
+
+	val, ok := clusterRoleBinding.Annotations[openShiftManagedRoleBindingAnnotation]
```
sttts, line 87: this check I would move into the main loop.

pkg/operator/rbaccontroller/rbaccontroller.go
```diff
@@ -0,0 +1,100 @@
...
+	"k8s.io/apiserver/pkg/authentication/user"
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/kubernetes"
+	listerrbacv1 "k8s.io/client-go/listers/rbac/v1"
+	"k8s.io/klog"
+)
+
+const (
+	// annotation used to indicate if unauthenticated user needs to be
+	// removed from the list of subjects for a cluster role binding
```
sttts, line 21: hmm, this is still very specific to the bug, not generic. Code comments should be generic, i.e. describe the mechanism, even if the mechanism is used to fix the bug.

pkg/operator/rbaccontroller/rbaccontroller.go
```diff
@@ -0,0 +1,100 @@
...
+	openShiftManagedRoleBindingAnnotation = "openshiftapiserver.rbac.authorization.openshift.io/managed"
+)
+
+// RBACController updates a cluster role binding after
+// removing unauthenticated user from the list of subjects
```
sttts, line 26: same as above

pkg/operator/rbaccontroller/rbaccontroller.go
```diff
@@ -0,0 +1,100 @@
...
+type RBACController struct {
+	kubeClient               kubernetes.Interface
+	clusterRoleBindingLister listerrbacv1.ClusterRoleBindingLister
+	eventRecorder            events.Recorder
+}
+
+func NewRBACController(kubeClient kubernetes.Interface, informerFactory informers.SharedInformerFactory, recorder events.Recorder) factory.Controller {
+
```
sttts, line 34: nit: no empty line

pkg/operator/rbaccontroller/rbaccontroller.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	return factory.New().
+		WithInformers(informerFactory.Rbac().V1().ClusterRoleBindings().Informer()).
+		WithSync(c.sync).
+		ToController("RBACController", c.eventRecorder)
+}
+
+func (c *RBACController) sync(ctx context.Context, syncContext factory.SyncContext) error {
+	clusterRoleBindingNames := []string{"discovery", "system:basic-user", "system:discovery", "system:openshift:discovery", "cluster-status-binding"}
```
sttts, line 48: comment what is in this list. Why these?

pkg/operator/rbaccontroller/rbaccontroller.go
```diff
@@ -0,0 +1,100 @@
...
+			errs = append(errs, err)
+		}
+	}
+
+	return errors.NewAggregate(errs)
+}
+
+// withoutUnauthenticatedSubject returns a new copy of given cluster role binding
+// without unauthenticated user in subjects
+func withoutUnauthenticatedSubject(clusterRoleBinding *rbacv1.ClusterRoleBinding) (*rbacv1.ClusterRoleBinding, bool) {
```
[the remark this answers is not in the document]
sttts: this name was my suggestion. How about `withUnauthenticatedSubjectRemoved`?

pkg/operator/rbaccontroller/rbaccontroller.go
```diff
@@ -0,0 +1,100 @@
...
+	if clusterRoleBinding == nil {
+		return nil, false
+	}
+
+	clusterRoleBindingCopy := clusterRoleBinding.DeepCopy()
```
sttts, line 89: nit: this deep copy is only necessary in the if-clause of the loop

pkg/operator/starter.go
sttts: Line 71 has the empty namespace. My assumption was that this mean cluster-wide resources?

pkg/operator/rbaccontroller/rbaccontroller.go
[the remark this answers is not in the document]
marun: To me, `remove` seems like a more apt description of intent. 'with/without' seems better suited to the name of a method that provided a filtered view of an instance or field of that instance, since the instance/field are naturally the target of 'remove'.  But nevermind, this is clearly an exercise in bikeshedding.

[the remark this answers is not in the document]
sttts: <programming-lang-bike-shed>removeXyz is imperative, withXyzRemove is functional</programming-lang-bike-shed>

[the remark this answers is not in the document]
vareti: changed function name to `withUnauthenticatedSubjectRemoved`

pkg/operator/rbaccontroller/rbaccontroller.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+func (c *RBACController) sync(ctx context.Context, syncContext factory.SyncContext) error {
+	clusterRoleBindingNames := []string{"discovery", "system:basic-user", "system:discovery", "system:openshift:discovery", "cluster-status-binding"}
+
+	var errs []error
+
+	for _, clusterRoleBindingName := range clusterRoleBindingNames {
+		clusterRoleBinding, err := c.clusterRoleBindingLister.Get(clusterRoleBindingName)
+		if err != nil {
```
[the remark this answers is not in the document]
vareti: Yes, that was my intent. Updated the implementation to what you suggested.

pkg/operator/rbaccontroller/rbaccontroller.go
vareti: gave another shot at it. please take a look.

pkg/operator/rbaccontroller/rbaccontroller.go
vareti: you are right. moved it inside if loop

pkg/operator/rbaccontroller/rbaccontroller.go
vareti: added a comment now

pkg/operator/rbaccontroller/rbaccontroller.go
vareti: done!

pkg/operator/rbaccontroller/rbaccontroller.go
```diff
@@ -0,0 +1,99 @@
...
+	"k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/apiserver/pkg/authentication/user"
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/kubernetes"
+	listerrbacv1 "k8s.io/client-go/listers/rbac/v1"
+	"k8s.io/klog"
+)
+
+const (
+	// openShiftManagedRBACAnnotation is the name of an annotation which prevents reconciliation if set to "false"
```
sttts, line 21: Doesn't it mean that the `openshiftapiserver.rbac.authorization.openshift.io/managed: true` means that it is owned by the operator, and therefore reconciled against the latest version? It will never be false. Or can and may a user set it to false?

vareti: Yes. It is owned and reconciled only when `openshiftapiserver.rbac.authorization.openshift.io/managed: true`. If this annotation is missing or when user set it to false, reconcile is skipped.

pkg/operator/starter.go
vareti: Yes, it means cluster-wide resources.

sttts: Can't we use it instead of another informer factory?

vareti: Good point. Fixed it to re-use informer factory.

pkg/operator/rbaccontroller/rbaccontroller.go
```diff
@@ -0,0 +1,105 @@
...
+	"k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/apiserver/pkg/authentication/user"
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/kubernetes"
+	listerrbacv1 "k8s.io/client-go/listers/rbac/v1"
+	"k8s.io/klog"
+)
+
+const (
+	// If set to "true" for a cluster role binding, RBACController would remove unauthenticated
```
marun, line 21: As per a previous comment from @sttts, please indicate that this annotation is set by the operand e.g.

```
This annotation is set to a value of `true` on RBAC resources created by the openshift-apiserver.
```

pkg/operator/rbaccontroller/rbaccontroller.go
```diff
@@ -0,0 +1,105 @@
...
+			// Not every cluster role binding is present in every version of OpenShift.
+			// Downgrading the error to avoid log spamming
+			klog.V(4).Infof("fetching cluster role binding %s failed %v", clusterRoleBindingName, err)
+			continue
+		} else if err != nil {
+			errs = append(errs, err)
+			continue
+		}
+
+		if val, _ := clusterRoleBinding.Annotations[openShiftManagedRBACAnnotation]; val != "true" {
```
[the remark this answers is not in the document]
vareti: Got it! Updated.

pkg/operator/rbaccontroller/rbaccontroller.go
```diff
@@ -0,0 +1,105 @@
...
+	"k8s.io/klog"
+)
+
+const (
+	// If set to "true" for a cluster role binding, RBACController would remove unauthenticated
+	// subject from list of subjects
+	openShiftManagedRBACAnnotation = "openshiftapiserver.rbac.authorization.openshift.io/managed"
+)
+
+// RBACController reconciles RBAC resources created by openshift-apiserver
```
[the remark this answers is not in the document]
vareti: Done.

pkg/operator/rbaccontroller/rbaccontroller.go
```diff
@@ -0,0 +1,105 @@
...
+
+	return factory.New().
+		WithInformers(informerFactory.Rbac().V1().ClusterRoleBindings().Informer()).
+		WithSync(c.sync).
+		ToController("RBACController", c.eventRecorder)
+}
+
+func (c *RBACController) sync(ctx context.Context, syncContext factory.SyncContext) error {
+	// cluster role binding names that granted unauthenticated user access to "/" endpoint
+	clusterRoleBindingNames := []string{"discovery", "basic-users", "system:openshift:discovery",
```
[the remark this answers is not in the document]
vareti: Updated the comment.

pkg/operator/rbaccontroller/rbaccontroller.go
```diff
@@ -0,0 +1,105 @@
...
+			errs = append(errs, err)
+		}
+	}
+
+	return errors.NewAggregate(errs)
+}
+
+// withUnauthenticatedSubjectRemoved returns a new copy of given cluster role binding
+// without unauthenticated user in subjects
+func withUnauthenticatedSubjectRemoved(clusterRoleBinding *rbacv1.ClusterRoleBinding) (*rbacv1.ClusterRoleBinding, bool) {
```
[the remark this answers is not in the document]
vareti: simplified it now.

pkg/operator/rbaccontroller/rbaccontroller.go
```diff
@@ -0,0 +1,105 @@
...
+		eventRecorder:            recorder,
+	}
+
+	return factory.New().
+		WithInformers(informerFactory.Rbac().V1().ClusterRoleBindings().Informer()).
+		WithSync(c.sync).
+		ToController("RBACController", c.eventRecorder)
+}
+
+func (c *RBACController) sync(ctx context.Context, syncContext factory.SyncContext) error {
```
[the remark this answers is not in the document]
vareti: :smile:

pkg/operator/rbaccontroller/rbaccontroller.go
```diff
@@ -0,0 +1,105 @@
...
+		_, err = c.kubeClient.RbacV1().ClusterRoleBindings().Update(ctx, updatedCopy, metav1.UpdateOptions{})
+		if err != nil {
+			errs = append(errs, err)
+		}
+	}
+
+	return errors.NewAggregate(errs)
+}
+
+// withUnauthenticatedSubjectRemoved returns a new copy of given cluster role binding
```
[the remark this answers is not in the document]
vareti: Done.

pkg/operator/rbaccontroller/rbaccontroller.go
vareti: I see what is missing! updated the comment now.

vareti:
@marun Thanks for the detailed review :+1: Addressed your comments! PTAL

pkg/operator/rbaccontroller/rbaccontroller.go
```diff
@@ -0,0 +1,106 @@
...
+			errs = append(errs, err)
+		}
+	}
+
+	return errors.NewAggregate(errs)
+}
+
+// withUnauthenticatedSubjectRemoved returns a new copy of given cluster role binding without unauthenticated user in subjects.
+func withUnauthenticatedSubjectRemoved(clusterRoleBinding *rbacv1.ClusterRoleBinding) *rbacv1.ClusterRoleBinding {
+	if clusterRoleBinding == nil {
```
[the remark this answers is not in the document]
vareti: I left it as a safe guard so that code does not panic when nil object is accidentally passed by a new caller.

pkg/operator/rbaccontroller/rbaccontroller.go
```diff
@@ -0,0 +1,106 @@
...
+
+const (
+	// This annotation is set to a value of "true" on RBAC resources created by the openshift-apiserver.
+	openShiftManagedRBACAnnotation = "openshiftapiserver.rbac.authorization.openshift.io/managed"
+)
+
+// RBACController reconciles RBAC resources created by openshift-apiserver.
+//
+// If openshiftManagedRBACAnnotation is set to "true" for a cluster role binding,
+// RBACController would remove unauthenticated subject from list of subjects.
```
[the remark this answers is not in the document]
vareti: Done.

vareti:
> Is it your intent to reconcile on every change to every cluster role binding, or is that just an accident of implementation?

It is just an accident of implementation. I did not want to diverge from factory controller implementation. Building a controller from scratch would result in repeated code?

pkg/operator/rbaccontroller/rbaccontroller.go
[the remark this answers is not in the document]
marun: (No action required, just food for thought)

There's this principle I really like - YAGNI (you ain't gonna need it). Err on the side of doing less if it isn't required today. It may be that more will never be required, and if it is, you can do it in the future instead of now. 

Another thought - have the caller dereference the pointer. The return value is a copy after all - the function doesn't have to modify it.

pkg/operator/rbaccontroller/rbaccontroller.go
```diff
@@ -0,0 +1,145 @@
...
+
+const (
+	// This annotation is set to a value of "true" on RBAC resources created by the openshift-apiserver.
+	openShiftManagedRBACAnnotation = "openshiftapiserver.rbac.authorization.openshift.io/managed"
+)
+
+var (
+	// The named cluster role bindings are created by openshift-apiserver
+	// and in previous releases may have included the unauthenticated subject.
+	clusterRoleBindingNames = []string{"discovery", "basic-users", "system:openshift:discovery",
```
marun, line 29: Use a set to simplify identifying a match.

pkg/operator/rbaccontroller/rbaccontroller.go
```diff
@@ -0,0 +1,145 @@
...
+		kubeClient:               kubeClient,
+		clusterRoleBindingLister: informerFactory.Rbac().V1().ClusterRoleBindings().Lister(),
+		eventRecorder:            recorder,
+	}
+
+	syncCtx := factory.NewSyncContext("RBACController", c.eventRecorder)
+
+	informer := informerFactory.Rbac().V1().ClusterRoleBindings().Informer()
+	informer.AddEventHandler(cache.FilteringResourceEventHandler{
+		FilterFunc: hasMatchingNames,
```
marun, line 56: Filter on the annotation as well as the name.

pkg/operator/rbaccontroller/rbaccontroller.go
```diff
@@ -0,0 +1,145 @@
...
+		WithSyncContext(syncCtx).
+		WithSync(c.sync).
+		ToController("", nil)
+}
+
+// sync ensures that cluster role bindings created by openshift-apiserver do not have the unauthenticated subject.
+func (c *RBACController) sync(ctx context.Context, syncContext factory.SyncContext) error {
+	var errs []error
+
+	for _, clusterRoleBindingName := range clusterRoleBindingNames {
```
marun, line 84: With the name and annotation filtering in place, it is no longer necessary to iterate over target names or check for the annotation. `syncContext` provides a key identifying the CRB that can be used to retrieve the resource from the informer cache.

pkg/operator/rbaccontroller/rbaccontroller.go
vareti: Done.

pkg/operator/rbaccontroller/rbaccontroller.go
vareti: now we filter on the annotations as well as the name.

pkg/operator/rbaccontroller/rbaccontroller.go
vareti: I tried using informers and filters but not sure if I am using them the proper way. So, It looks like correct :smile: 

I updated the code based on your comments. Please take a look!

pkg/operator/rbaccontroller/rbaccontroller.go
sttts: RBACReconcileController would be more specific.

pkg/operator/rbaccontroller/rbaccontroller.go
```diff
@@ -0,0 +1,126 @@
...
+
+const (
+	// This annotation is set to a value of "true" on RBAC resources created by the openshift-apiserver.
+	openShiftManagedRBACAnnotation = "openshiftapiserver.rbac.authorization.openshift.io/managed"
+)
+
+var (
+	// The named cluster role bindings are created by openshift-apiserver
+	// and in previous releases may have included the unauthenticated subject.
+	clusterRoleBindingNames = sets.NewString("discovery", "basic-users", "system:openshift:discovery",
```
sttts, line 27: what is this for? I thought we have the annotation to know the bindings we are interested in here.

pkg/operator/rbaccontroller/rbaccontroller.go
```diff
@@ -0,0 +1,126 @@
...
+	if updatedCopy == nil {
+		return nil
+	}
+
+	_, err = c.kubeClient.RbacV1().ClusterRoleBindings().Update(ctx, updatedCopy, metav1.UpdateOptions{})
+	return err
+}
+
+// withUnauthenticatedSubjectRemoved returns a new copy of given cluster role binding without unauthenticated user in subjects.
+func withUnauthenticatedSubjectRemoved(clusterRoleBinding *rbacv1.ClusterRoleBinding) *rbacv1.ClusterRoleBinding {
```
sttts, line 100: returning nil for a noop is not typical Golang style. Add a "changed bool".

pkg/operator/rbaccontroller/rbaccontroller.go
```diff
@@ -0,0 +1,126 @@
...
+			return clusterRoleBindingCopy
+		}
+	}
+
+	return nil
+}
+
+// hasMatchingAttributes returns true if the obj name equals one of
+// clusterRoleBindingNames and it's RBAC annotation is set to "true".
+func hasMatchingAttributes(obj interface{}) bool {
```
sttts, line 115: be specific. What does this filter for?

sttts:
/close

We switched to a doc approach as changing this automatically might break existing clusters.

PR #354, opened 2020-04-16
Bug 1827050: Update to use revised ApplyDeployment

marun:
The new method hashes the spec of the deployment which removes the setting an annotation containing the replica count.

/cc @deads2k @stlaz


marun:
Updated to use merged library-go change, PTAL.

marun:
/retitle Bug 1827050: Update to use revised ApplyDeployment

marun:
Rebased, PTAL

PR #355, opened 2020-04-16
Bug 1828835: remove duplicate migration informer start

vareti:
`migrationInformer.Start(ctx.Done())` is repeated twice and one instance of it can be removed


vareti:
/retitle Bug 1828835: remove duplicate migration informer start

PR #360, opened 2020-05-06
Bug 1832220: operator: add openshift-etcd-operator namespace to related objects

mfojtik:
This will cause insights operator to capture events from etcd operator namespace in case openshift apiserver operator is degraded.


sttts:
lgtm

PR #362, opened 2020-05-08
Bug 1795163: openshift-apiserver operator not available when used for single node cluster (CRC) 

p0lyn0mial:
reverts waiting for `requestheader-client-ca-file` in `extension-apiserver-authentication` before rolling out a new version.

it is replaced by https://github.com/openshift/kubernetes-apiserver/pull/10 which essentially shifts the logic to API Servers. From now on a server watches and reacts to changes in the config map in real-time.

### DIFF
--- a/pkg/operator/starter.go
+++ b/pkg/operator/starter.go
@@ -190,7 +190,6 @@ func RunOperator(ctx context.Context, controllerConfig *controllercmd.Controller
 		configClient.ConfigV1().ClusterOperators(),
 		versionRecorder,
 		kubeInformersForNamespaces,
-		kubeInformersForNamespaces.InformersFor(metav1.NamespaceSystem).Core().V1().ConfigMaps().Informer(),
 		operatorConfigInformers.Operator().V1().OpenShiftAPIServers().Informer(),
 		configInformers.Config().V1().Images().Informer(),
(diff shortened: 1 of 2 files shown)


pkg/operator/starter.go
```diff
@@ -190,7 +190,6 @@ func RunOperator(ctx context.Context, controllerConfig *controllercmd.Controller
 		configClient.ConfigV1().ClusterOperators(),
 		versionRecorder,
 		kubeInformersForNamespaces,
-		kubeInformersForNamespaces.InformersFor(metav1.NamespaceSystem).Core().V1().ConfigMaps().Informer(),
```
p0lyn0mial, line 193: I can't remove `metav1.NamespaceSystem` from `v1helpers.NewKubeInformersForNamespaces` because it's being used in https://github.com/openshift/cluster-openshift-apiserver-operator/blob/master/vendor/github.com/openshift/library-go/pkg/operator/apiserver/controllerset/apiservercontrollerset.go#L235

this should be changed in `master` branch since `Nodes` is a non-namespaced resources.

p0lyn0mial:
the original commit was https://github.com/openshift/cluster-openshift-apiserver-operator/commit/33ce8dfee2901f5698c22c3419fa41cbea075200

PR #364, opened 2020-05-15
Bug 1836927: Prefer etcd endpoints configmap for storage URL discovery

ironcladlou:
xref https://github.com/openshift/cluster-etcd-operator/pull/354

### DIFF
--- a/pkg/operator/configobservation/configobservercontroller/observe_config_controller.go
+++ b/pkg/operator/configobservation/configobservercontroller/observe_config_controller.go
@@ -16,7 +16,7 @@ import (
 	"github.com/openshift/library-go/pkg/operator/v1helpers"
 
 	"github.com/openshift/cluster-openshift-apiserver-operator/pkg/operator/configobservation"
-	"github.com/openshift/cluster-openshift-apiserver-operator/pkg/operator/configobservation/etcdobserver"
+	"github.com/openshift/cluster-openshift-apiserver-operator/pkg/operator/configobservation/etcdendpoints"
 	"github.com/openshift/cluster-openshift-apiserver-operator/pkg/operator/configobservation/images"
 	"github.com/openshift/cluster-openshift-apiserver-operator/pkg/operator/configobservation/ingresses"
 	"github.com/openshift/cluster-openshift-apiserver-operator/pkg/operator/configobservation/project"
@@ -44,6 +44,7 @@ func NewConfigObserver(
 			ProxyLister_:        configInformers.Config().V1().Proxies().Lister(),
 			IngressConfigLister: configInformers.Config().V1().Ingresses().Lister(),
 			EndpointsLister:     kubeInformersForEtcdNamespace.Core().V1().Endpoints().Lister(),
+			ConfigmapLister:     kubeInformersForEtcdNamespace.Core().V1().ConfigMaps().Lister(),
 			SecretLister_:       kubeInformers.Core().V1().Secrets().Lister(),
 			PreRunCachesSynced: []cache.InformerSynced{
 				operatorConfigInformers.Operator().V1().OpenShiftAPIServers().Informer().HasSynced,
@@ -54,6 +55,7 @@ func NewConfigObserver(
 				configInformers.Config().V1().Ingresses().Informer().HasSynced,
 				kubeInformersForEtcdNamespace.Core().V1().Endpoints().Informer().HasSynced,
 				kubeInformers.Core().V1().Secrets().Informer().HasSynced,
+				kubeInformersForEtcdNamespace.Core().V1().ConfigMaps().Informer().HasSynced,
 			},
 		},
 		[]factory.Informer{operatorConfigInformers.Operator().V1().OpenShiftAPIServers().Informer()},
@@ -61,7 +63,7 @@ func NewConfigObserver(
 		images.ObserveExternalRegistryHostnames,
 		images.ObserveAllowedRegistriesForImport,
 		ingresses.ObserveIngressDomain,
-		etcdobserver.ObserveStorageURLs,
+		etcdendpoints.ObserveStorageURLs,
 		libgoapiserver.ObserveTLSSecurityProfile,
 		project.ObserveProjectRequestMessage,
 		project.ObserveProjectRequestTemplateName,
--- a/pkg/operator/configobservation/etcdendpoints/observe_etcd_endpoints.go
+++ b/pkg/operator/configobservation/etcdendpoints/observe_etcd_endpoints.go
@@ -1,0 +1,106 @@
+package etcdendpoints
+
+import (
+	"fmt"
+	"net"
+	"reflect"
+	"sort"
+	"strings"
+
+	"github.com/openshift/library-go/pkg/operator/configobserver"
+	"github.com/openshift/library-go/pkg/operator/events"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+
+	"github.com/openshift/cluster-openshift-apiserver-operator/pkg/operator/configobservation"
+	endpointsobserver "github.com/openshift/cluster-openshift-apiserver-operator/pkg/operator/configobservation/etcdobserver"
+)
+
+const (
+	etcdEndpointNamespace = "openshift-etcd"
+	etcdEndpointName      = "etcd-endpoints"
+)
+
+var fallbackObserver configobserver.ObserveConfigFunc = endpointsobserver.ObserveStorageURLs
+
+// ObserveStorageURLs observes the storage config URLs. If there is a problem observing the current storage config URLs,
+// then the previously observed storage config URLs will be re-used.
+func ObserveStorageURLs(genericListers configobserver.Listers, recorder events.Recorder, currentConfig map[string]interface{}) (map[string]interface{}, []error) {
+	listers := genericListers.(configobservation.Listers)
+	storageConfigURLsPath := []string{"storageConfig", "urls"}
+	var errs []error
+
+	previouslyObservedConfig := map[string]interface{}{}
+	currentEtcdURLs, found, err := unstructured.NestedStringSlice(currentConfig, storageConfigURLsPath...)
+	if err != nil {
+		errs = append(errs, err)
+	}
+	if found {
+		if err := unstructured.SetNestedStringSlice(previouslyObservedConfig, currentEtcdURLs, storageConfigURLsPath...); err != nil {
+			errs = append(errs, err)
+		}
+	}
+
+	var etcdURLs []string
+	etcdEndpoints, err := listers.ConfigmapLister.ConfigMaps(etcdEndpointNamespace).Get(etcdEndpointName)
+	if errors.IsNotFound(err) {
+		// In clusters prior to 4.5, fall back to reading the old host-etcd-2 endpoint
+		// resource, if possible. In 4.6 we can assume consumers have been updated to
+		// use the configmap, delete the fallback code, and throw an error if the
+		// configmap doesn't exist.
+		observedConfig, fallbackErrors := fallbackObserver(listers, recorder, currentConfig)
+		if len(fallbackErrors) > 0 {
+			errs = append(errs, fallbackErrors...)
+			return previouslyObservedConfig, append(errs, fmt.Errorf("configmap %s/%s not found, and fallback observer failed", etcdEndpointNamespace, etcdEndpointName))
+		}
+		return observedConfig, errs
+	}
+	if err != nil {
+		recorder.Warningf("ObserveStorageFailed", "Error getting %s/%s configmap: %v", etcdEndpointNamespace, etcdEndpointName, err)
+		return previouslyObservedConfig, append(errs, err)
+	}
+
+	// note: etcd bootstrap should never be added to the in-cluster kube-apiserver
+	// this can result in some early pods crashlooping, but ensures that we never contact the bootstrap machine from
+	// the in-cluster kube-apiserver so we can safely teardown out of order.
+
+	for k := range etcdEndpoints.Data {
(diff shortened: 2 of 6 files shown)


pkg/operator/configobservation/configobservercontroller/observe_config_controller.go
```diff
@@ -62,6 +64,7 @@ func NewConfigObserver(
 		images.ObserveAllowedRegistriesForImport,
 		ingresses.ObserveIngressDomain,
 		etcdobserver.ObserveStorageURLs,
+		etcdendpoints.ObserveStorageURLs,
```
sanchezl, line 67: Please add test that combines the expected output of both just to be sure they don't step on each other.
I don't remember the exact rules `merge.Merge` (from https://github.com/openshift/cluster-kube-apiserver-operator/blob/22501ab373f961b9b5519803dc031830dd02defb/vendor/github.com/openshift/library-go/pkg/operator/configobserver/config_observer_controller.go#L92) uses.

ironcladlou: https://github.com/openshift/cluster-kube-apiserver-operator/pull/859 Over in the sibling PR I refactored to an approach that keeps the observer code separate but chooses the impl using a simple adapter function up front, take a look and comment over there and whatever approach we land on I'll port to this PR

pkg/operator/configobservation/etcdendpoints/observe_etcd_endpoints.go
```diff
@@ -0,0 +1,110 @@
...
+// etcd-endpoints configmap, and will fall back to reading the endpoint
+// resource only if the configmap doesn't yet exist.
+func BackwardsCompatibleObserveStorageURLs(genericListers configobserver.Listers, recorder events.Recorder, currentConfig map[string]interface{}) (map[string]interface{}, []error) {
+	listers := genericListers.(configobservation.Listers)
+	if _, err := listers.ConfigmapLister.ConfigMaps(etcdEndpointNamespace).Get(etcdEndpointName); err == nil {
+		return ObserveStorageURLs(genericListers, recorder, currentConfig)
+	} else if !errors.IsNotFound(err) {
+		return nil, []error{err}
+	}
+	recorder.Warningf("ObserveStorageFallback", "%s/%s configmap not found, falling back to host-etcd-2 endpoint", etcdEndpointNamespace, etcdEndpointName)
```
ironcladlou, line 33: @sanchezl @deads2k this does generate a lot of event noise before the corresponding etcd-operator transition (3212 events in the e2e-aws run), but that should of course taper off and cease entirely when etcd-operator is updated and generates the configmap.

I'm not sure if the event volume is concerning on its own, or whether the warning provides enough value to justify it. I'll keep it for now, let me know what you think.

ironcladlou:
@deads2k @sanchezl @hexfusion @retroflexer this is ready for review.

pkg/operator/configobservation/etcdendpoints/observe_etcd_endpoints.go
deads2k: > I'm not sure if the event volume is concerning on its own, or whether the warning provides enough value to justify it. I'll keep it for now, let me know what you think.

it does not.  remove it please

pkg/operator/configobservation/etcdendpoints/observe_etcd_endpoints.go
```diff
@@ -0,0 +1,110 @@
...
+
+// BackwardsCompatibleObserveStorageURLs will prefer to observe the new
+// etcd-endpoints configmap, and will fall back to reading the endpoint
+// resource only if the configmap doesn't yet exist.
+func BackwardsCompatibleObserveStorageURLs(genericListers configobserver.Listers, recorder events.Recorder, currentConfig map[string]interface{}) (map[string]interface{}, []error) {
+	listers := genericListers.(configobservation.Listers)
+	if _, err := listers.ConfigmapLister.ConfigMaps(etcdEndpointNamespace).Get(etcdEndpointName); err == nil {
+		return ObserveStorageURLs(genericListers, recorder, currentConfig)
+	} else if !errors.IsNotFound(err) {
+		return nil, []error{err}
```
deads2k, line 31: Return the previous value.  As I recall, doing otherwise can result in clearing our values.

pkg/operator/configobservation/etcdendpoints/observe_etcd_endpoints.go
```diff
@@ -0,0 +1,110 @@
...
+			etcdURLs = append(etcdURLs, fmt.Sprintf("https://%s:2379", ip))
+		} else {
+			etcdURLs = append(etcdURLs, fmt.Sprintf("https://[%s]:2379", ip))
+		}
+	}
+
+	if len(etcdURLs) == 0 {
+		emptyURLErr := fmt.Errorf("configmaps %s/%s: no etcd endpoint addresses found", etcdEndpointNamespace, etcdEndpointName)
+		recorder.Warning("ObserveStorageFailed", emptyURLErr.Error())
+		errs = append(errs, emptyURLErr)
```
deads2k, line 94: this was probably cargo-culted, but in this condition, please return the previous config and errors here.

pkg/operator/configobservation/configobservercontroller/observe_config_controller.go
```diff
@@ -44,6 +44,7 @@ func NewConfigObserver(
 			ProxyLister_:        configInformers.Config().V1().Proxies().Lister(),
 			IngressConfigLister: configInformers.Config().V1().Ingresses().Lister(),
 			EndpointsLister:     kubeInformersForEtcdNamespace.Core().V1().Endpoints().Lister(),
+			ConfigmapLister:     kubeInformersForEtcdNamespace.Core().V1().ConfigMaps().Lister(),
```
deads2k, line 47: add to preruncachessynced

deads2k, line 47: configmaps should be added to the informers above so that we trigger on it.

ironcladlou: Not sure what this one is referring to — link to where you are talking about adding?

ironcladlou: Do you mean adding it here?

https://github.com/openshift/cluster-openshift-apiserver-operator/blob/master/pkg/operator/configobservation/configobservercontroller/observe_config_controller.go#L59

If so, why aren't all the other informers in the lister and sync list also in the cited informer list?

ironcladlou: Fixed

pkg/operator/configobservation/etcdendpoints/observe_etcd_endpoints.go
ironcladlou: Fixed

pkg/operator/configobservation/etcdendpoints/observe_etcd_endpoints.go
ironcladlou: Fixed

pkg/operator/configobservation/etcdendpoints/observe_etcd_endpoints.go
ironcladlou: Fixed

pkg/operator/configobservation/configobservercontroller/observe_config_controller.go
ironcladlou: Added the configmap informer to the list, but the question remains re: the other informers

pkg/operator/configobservation/etcdendpoints/observe_etcd_endpoints.go
ironcladlou: This is actually codified in unit tests, which are now breaking

ironcladlou: I restored the original behavior because there are two explicit test cases around it and I'm not confident in changing it with this PR. If you and @hexfusion and @retroflexer are convinced the current behavior is wrong, let's describe the bug and do a followup.

pkg/operator/configobservation/configobservercontroller/observe_config_controller.go
sanchezl: @deads2k / @ironcladlou The `library-go` `ConfigObserver`  [extracts the needed Informers from the Listers](https://github.com/openshift/cluster-openshift-apiserver-operator/blob/e1f97fb02db81f8f69b1750e18d55156c885ee7a/vendor/github.com/openshift/library-go/pkg/operator/configobserver/config_observer_controller.go#L65).

pkg/operator/configobservation/etcdendpoints/observe_etcd_endpoints.go
```diff
@@ -0,0 +1,109 @@
...
+		if err := unstructured.SetNestedStringSlice(previouslyObservedConfig, currentEtcdURLs, storageConfigURLsPath...); err != nil {
+			errs = append(errs, err)
+		}
+	}
+
+	var etcdURLs []string
+	etcdEndpoints, err := listers.ConfigmapLister.ConfigMaps(etcdEndpointNamespace).Get(etcdEndpointName)
+	if errors.IsNotFound(err) {
+		recorder.Warningf("ObserveStorageFailed", "Required %s/%s configmap not found", etcdEndpointNamespace, etcdEndpointName)
+		return previouslyObservedConfig, append(errs, fmt.Errorf("configmaps/%s.%s: not found", etcdEndpointName, etcdEndpointNamespace))
```
sanchezl, line 58:
```suggestion
		recorder.Warningf("ObserveStorageFailed", "Required %s/%s configmap not found; falling back to deprecated endpoint.", etcdEndpointNamespace, etcdEndpointName)
		errs = append(errs, fmt.Errorf("configmaps/%s.%s: not found", etcdEndpointName, etcdEndpointNamespace))
		observedConfig, observerErrs := endpointsobserver.ObserveStorageURLs(genericListers, recorder, currentConfig)
		return observerConfig, append(errs, observerErrs...)
```
I still think it would be best to call the "legacy" observer func here instead of using `BackwardsCompatibleObserveStorageURLs`.  The err!=nil when looking for configmap case below returns slightly different config & errs than in `BackwardsCompatibleObserveStorageURLs`.

ironcladlou: >I still think it would be best to call the "legacy" observer func here instead of using BackwardsCompatibleObserveStorageURLs. The err!=nil when looking for configmap case below returns slightly different config & errs than in BackwardsCompatibleObserveStorageURLs.

The advantage of using `BackwardsCompatibleObserveStorageURLs` is that the actual observation code for both old and new remain orthogonal. If we make the decision in ObserveStorageURLs, now the old and new paths are coupled, and we'll even need to expand the test coverage of the new one to incorporate awareness of the old one.

However, I see your concern about the variation in return semantics in the configmap missing case. To address both our concerns: what if we changed `BackwardsCompatibleObserveStorageURLs` so that when the configmap lookup fails for a non-missing error, we do the same `previouslyObservedConfig` logic the observer functions perform so that the output is consistent across all three methods (adapter, old and new observers)?

ironcladlou: @sanchezl

ironcladlou: Refactored to incorporate feedback

pkg/operator/configobservation/configobservercontroller/observe_config_controller.go
ironcladlou: Excellent, thanks

ironcladlou:
@hexfusion @deads2k @sanchezl this is once again ready for review, PTAL

pkg/operator/configobservation/etcdendpoints/observe_etcd_endpoints.go
ironcladlou: I actually decided to fix this after all, because it exposed a bug whereby the localhost was erroneously being appended regardless of whether endpoints were discovered.

pkg/operator/configobservation/configobservercontroller/observe_config_controller.go
```diff
@@ -54,14 +55,15 @@ func NewConfigObserver(
 				configInformers.Config().V1().Ingresses().Informer().HasSynced,
 				kubeInformersForEtcdNamespace.Core().V1().Endpoints().Informer().HasSynced,
 				kubeInformers.Core().V1().Secrets().Informer().HasSynced,
+				kubeInformersForEtcdNamespace.Core().V1().ConfigMaps().Informer().HasSynced,
 			},
 		},
 		[]factory.Informer{operatorConfigInformers.Operator().V1().OpenShiftAPIServers().Informer()},
 		images.ObserveInternalRegistryHostname,
```
deads2k, line 62: add the configmap informer for the etcd namespace to the list of informers one line up

deads2k:
missing trigger.  It's not technically worse than before, but it's not good.

/lgtm

pkg/operator/configobservation/configobservercontroller/observe_config_controller.go
ironcladlou: https://github.com/openshift/cluster-openshift-apiserver-operator/pull/364#discussion_r427378029

looks like library-go does this automatically? if so the only informers i'd expect to pass to the constructor are those which aren't already in the listers

Or I totally misunderstood...

ironcladlou:
I'm not yet convinced there isn't some regression here. Need to see the latest test results.

ironcladlou:
Missing sort on the URL list was causing endless churn rolling out false diffs

PR #367, opened 2020-07-02
enable connectivity checks

sanchezl:


pkg/operator/connectivitycheckcontroller/connectivity_check_controller.go
```diff
@@ -0,0 +1,252 @@
...
+			TargetEndpoint: address,
+		},
+	}
+	for _, option := range options {
+		option(check)
+	}
+	return check
+}
+
+func withTlsClientCert(secretName string) func(*v1alpha1.PodNetworkConnectivityCheck) {
```
deads2k, line 246: nit: we use TLS

deads2k:
/lgtm

I think I looked at this before.

pkg/operator/connectivitycheckcontroller/connectivity_check_controller.go
sanchezl: @deads2k  https://github.com/openshift/cluster-openshift-apiserver-operator/pull/368

PR #370, opened 2020-07-20
enable scraping of check-endpoints metrics for openshift-apiserver

sanchezl:


manifests/0000_90_openshift-apiserver-operator_05_check-endpoints_servicemonitor.yaml
```diff
@@ -0,0 +1,19 @@
...
+  name: openshift-apiserver-operator-check-endpoints
+  namespace: openshift-apiserver
+spec:
+  endpoints:
+  - bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
+    interval: 30s
+    port: check-endpoints
+    scheme: https
+    tlsConfig:
+      insecureSkipVerify: true
```
deads2k, line 13: this one should be able to use the service-ca.

Also, we should have an e2e test that prevents these from being insecure.

deads2k:
/lgtm

followup to fix the cert.

PR #371, opened 2020-07-20
use the etcd observers from library-go and updates to rc2

p0lyn0mial:
this PR requires https://github.com/openshift/library-go/pull/839 first. It moves already existing etcd observer to the library-go.


sttts:
/hold
/approve
Lgtm

PR #376, opened 2020-08-05
connectivitycheckcontroller: sync w/ kube-apiserver improvements

sanchezl:
* name some endpoints using node names to exploit dynamic check updates
* add checks for load balancers
* refactor to make it easier to use conectivitycheckcontroller from library-go in the future.


deads2k:
well, CI results are down. looks ok, but I can't verify the output.  This looks to be converging with what I wrote in kas

/lgtm

PR #378, opened 2020-08-07
move to klog/v2

sanchezl:
```sh-session
$ find . -type f -name "*.go" -print0 | xargs -0 --max-procs=0 --max-args=1 sed --in-place --expression 's|"k8s.io/klog"|"k8s.io/klog/v2"|g' 
$ go mod vendor && go mod tidy
```


go.mod
```diff
@@ -7,6 +7,7 @@ require (
 	github.com/ghodss/yaml v1.0.0
 	github.com/go-bindata/go-bindata v3.1.2+incompatible
 	github.com/gonum/graph v0.0.0-20170401004347-50b27dea7ebb
+	github.com/imdario/mergo v0.3.7
```
deads2k, line 10: this loosk wrong

sanchezl: Cleaned up.

PR #381, opened 2020-08-24
bug 1871890: add current profile annotations to CVO manifests

deads2k:
This is matches https://github.com/openshift/enhancements/pull/414 and doesn't change existing behavior


soltysh (approved):
/lgtm
/hold
until https://github.com/openshift/cluster-openshift-apiserver-operator/pull/383 lands

PR #392, opened 2020-09-15
Bug 1878648: oauth-apiserver: audit log oauthaccesstokens if fresh 4.6 install

sttts:
Counterpart to https://github.com/openshift/cluster-authentication-operator/pull/324.

The annotation `oauth-apiserver.openshift.io/secure-token-storage: true` set on the config/v1 apiserver object make openshift-apiserver deeply log oauth tokens, noth access and authorize (if openshift-apiserver is responsible for this; in fresh 4.6 install it is not, but oauth-apiserver is). 

This requires that there are no non-sha256 tokens in the system. For 4.6 installs this is guaranteed. 

After downgrade to 4.5 there is a controller removing the annotation. 

After upgrade from 4.5, the annotation is not set. The admin can set it manually after having made sure that no non-sha256 tokens exist.

For upgrade to 4.7, non-sha256 tokens must not exist. If they do, the upgrade will be blocked by the auth operator (to be implemented).

### DIFF
--- a/pkg/operator/starter.go
+++ b/pkg/operator/starter.go
@@ -280,6 +280,22 @@ func RunOperator(ctx context.Context, controllerConfig *controllercmd.Controller
 		return err
 	}
 
+	// TODO(4.7): switch over to secure access-token logging by default and delete old non-sha256 tokens
+	auditPolicyPathGetterWithAccessTokenLogs := func(profile string) (string, error) {
+		apiServerConfig, err := configInformers.Config().V1().APIServers().Lister().Get("cluster")
+		if errors.IsNotFound(err) {
+			return auditPolicyPahGetter(profile)
+		}
+		if err != nil {
+			return "", err
+		}
+
+		if apiServerConfig.Annotations["oauth-apiserver.openshift.io/secure-token-storage"] == "true" {
+			return auditPolicyPahGetter("secure-oauth-storage-" + profile)
+		}
+		return auditPolicyPahGetter(profile)
+	}
+
 	configObserver := configobservercontroller.NewConfigObserver(
 		kubeInformersForNamespaces.InformersFor(operatorclient.TargetNamespace),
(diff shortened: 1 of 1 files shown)


pkg/operator/starter.go
```diff
@@ -280,14 +280,27 @@ func RunOperator(ctx context.Context, controllerConfig *controllercmd.Controller
 		return err
 	}
 
+	apiServerConfig, err := configClient.ConfigV1().APIServers().Get(ctx, "cluster", metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+
+	// TODO(4.7): switch over to secure access-token logging by default and delete old non-sha256 tokens
```
tkashem, line 288: @sttts maybe we should create a BZ for this and set target release to 4.7 so we can track it?

pkg/operator/starter.go
```diff
@@ -280,14 +280,27 @@ func RunOperator(ctx context.Context, controllerConfig *controllercmd.Controller
...
+	auditPolicyPathGetterWithAccessTokenLogs := func(profile string) (string, error) {
+		if apiServerConfig.Annotations["oauth-apiserver.openshift.io/secure-oauth-token-storage"] == "true" {
```
tkashem, line 290: what if the user removes the annotation from the APIServer object or set it to false? The `apiServerConfig` object is stale once it is changed by the user. May be I am misunderstanding the flow here?

pkg/operator/starter.go
tkashem: @sttts I created this BZ to track this: https://bugzilla.redhat.com/show_bug.cgi?id=1879182

sttts: :+1:

pkg/operator/starter.go
sttts: it will now redeploy with the old policy.

PR #397, opened 2020-09-29
Restrict log permissions

JAORMX:
While the current log directory permissions are restrictive and correct,
the file permissions are too permissive and it raises flags on
evaluations of the deployment. Let's instead change the permissions to
0600 which is more appropriate for these types of logs.

Note that while this is a superficial fix, it is also meant to address upgrades
(on an upgrade scenario, the permissions of old and rotated log files will be
fixed).

### DIFF
--- a/bindata/v3.11.0/openshift-apiserver/deploy.yaml
+++ b/bindata/v3.11.0/openshift-apiserver/deploy.yaml
@@ -46,7 +46,7 @@ spec:
           terminationMessagePolicy: FallbackToLogsOnError
           image: ${IMAGE}
           imagePullPolicy: IfNotPresent
-          command: ['sh', '-c', 'chmod 0700 /var/log/openshift-apiserver']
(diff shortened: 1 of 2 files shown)


JAORMX:
cc @stlaz

stlaz:
/hold
This is weird. Why not fix it in the component itself?

JAORMX:
> /hold
> This is weird. Why not fix it in the component itself?

In case of upgrades, this would fix the permissions of old logs too.

stlaz:
/hold cancel
That's alright, then

sttts:
This misses logs after log rotation.

JAORMX:
@sttts logs after log rotation inherit the initial permissions (at least that's how the underlying library is supposed to work)

JAORMX:
@sttts for reference, this attempts to fix the default in the underlying library https://github.com/natefinch/lumberjack/pull/112
and this attempts to change the permissions in kube: https://github.com/kubernetes/kubernetes/pull/95387

PR #401, opened 2020-10-12
no need to wait for kube-apiserver operator

deads2k:
this will increase the speed at which we get installed.

/hold

this is an illegal vendor change, but we should see what happens here.


deads2k:
going through the real changes now.

PR #404, opened 2020-10-13
bug 1888028: bump library-go to stop waiting on kube-apiserver availability

deads2k:


deads2k:
/hold

bump isn't real. just here so bug doesn't go modified.

deads2k:
/hold cancel

still the wrong pr

hexfusion (commented):
/lgtm

PR #417, opened 2020-11-23
don't claim control over the oauth and user API

stlaz:
Removes the oauth and user APIs from the list of
apiservices.

/assign @p0lyn0mial 

----

Seems straightforward, not sure if there are prereqs, but I think we should be fine?

PR #419, opened 2020-11-26
improve e2e tests

p0lyn0mial:
especially that we also run the encryption tests which require a few rollout and migration time.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -56,6 +56,7 @@ test-e2e-encryption-perf: test-unit
 
 .PHONY: test-e2e
 test-e2e: GO_TEST_PACKAGES :=./test/e2e/...
+test-e2e: GO_TEST_FLAGS += -timeout 1h
 test-e2e: test-unit
 
 # Configure the 'telepresence' target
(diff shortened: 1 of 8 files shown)


p0lyn0mial:
should unblock - https://github.com/openshift/cluster-openshift-apiserver-operator/pull/417

p0lyn0mial:
we observed frequent timeouts (10m) in #417.
the must-gather suggest that encryption was successful. we suspect that the default timeout is too low.

p0lyn0mial:
it looks like `TestRedeployOnConfigChange` (https://github.com/openshift/cluster-openshift-apiserver-operator/blob/master/test/e2e/operator_test.go#L34) test changes a config map that triggers a new revision and doesn't wait until all pods are installed. Not ideal for the encryption tests that need to wait longer.

p0lyn0mial:
> it looks like TestRedeployOnConfigChange (https://github.com/openshift/cluster-openshift-apiserver-operator/blob/master/test/e2e/operator_test.go#L34) test changes a config map that triggers a new revision and doesn't wait until all pods are installed. Not ideal for the encryption tests that need to wait longer.

the second commit addresses that, should be moved to `library-go`

p0lyn0mial:
requires https://github.com/openshift/library-go/pull/949

PR #422, opened 2020-12-04
wait for storage urls before installing the api server

p0lyn0mial:


pkg/operator/workload/workload_openshiftapiserver_v311_00_sync.go
```diff
@@ -105,16 +106,33 @@ func NewOpenShiftAPIServerWorkload(
...
-		return false, nil
+		return false, errors.New("waiting for observed configuration to be available")
+	}
+
+	// specifying etcd servers list is mandatory, without it the pods will be crashlooping, so wait for it.
+	// TODO: once OAS moves to using args directly (not via the config)
+	//       we could have a helper function for checking mandatory args for oauth-apiserver and openshift-apiserver
+	//       alternatively we could change the generic workload ctrl to accept a list of common preconditions
+
+	var typedObservedConfig openshiftcontrolplanev1.OpenShiftAPIServerConfig
```
p0lyn0mial, line 128: it is what we have in the config, if we don't like it we could unmarshal it to a map and get only what we need or read it from a serialized object

PR #440, opened 2021-03-30
expose pod information to oas container via env variables

p0lyn0mial:
the pod name and the namespace are used by the termination code to properly record events.

before the change the termination events were recorded to the default namespace and weren't associated with any pod.
after the change, the termination evens are recorded in the openshift-apiserver namespace and are bound to oas pods.

### DIFF
--- a/bindata/v3.11.0/openshift-apiserver/deploy.yaml
+++ b/bindata/v3.11.0/openshift-apiserver/deploy.yaml
@@ -73,6 +73,15 @@ spec:
           requests:
             memory: 200Mi
             cpu: 100m
+        env:
+          - name: POD_NAME
+            valueFrom:
+              fieldRef:
+                fieldPath: metadata.name
+          - name: POD_NAMESPACE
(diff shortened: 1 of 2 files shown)


p0lyn0mial:
I'm going to check `oauth-apiserver` as well.

p0lyn0mial:
> for the record TerminationMinimalShutdownDuration is set to 0 for both oas and oauth-apiserver. This seems to be ok since we already have terminationGracePeriodSeconds (70s) set.

this is wrong! `terminationGracePeriodSeconds` is how long kubelet waits before it terminates oas. It doesn't guarantee that it will always wait for that long.

That means our aggregated APIs are not terminated gracefully. 
According to Clayton, things on pod / service network need to wait 15-20s. That means we should set  `shutdown-delay-duration` to `15s`. Thanks!

p0lyn0mial:
I'm going to open follow-up PRs for `shutdown-delay-duration` next week, after my holidays.

PR #441, opened 2021-03-31
Bug 1943442: redeploy with maxSurge=1 in single-node to avoid unavailability

sttts:
- set maxSurge=1, maxUnavailability=0 in single-node on the deployment
- rotate audit logs start to avoid write conflicts

### DIFF
--- a/bindata/v3.11.0/openshift-apiserver/deploy.yaml
+++ b/bindata/v3.11.0/openshift-apiserver/deploy.yaml
@@ -7,17 +7,6 @@ metadata:
     app: openshift-apiserver
     apiserver: "true"
 spec:
-  # The number of replicas will be set in code to the number of master nodes.
-  strategy:
-    type: RollingUpdate
-    rollingUpdate:
-      # To ensure that only one pod at a time writes to the node's
-      # audit log, require the update strategy to proceed a node at a
-      # time. Only when a master node has its existing
-      # openshift-apiserver pod stopped will a new one be allowed to
-      # start.
-      maxUnavailable: 1
-      maxSurge: 0
   selector:
     matchLabels:
       # Need to vary the app label from that used by the legacy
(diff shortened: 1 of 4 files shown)


bindata/v3.11.0/openshift-apiserver/deploy.yaml
```diff
@@ -68,6 +57,9 @@ spec:
               echo "Copying system trust bundle"
               cp -f /var/run/configmaps/trusted-ca-bundle/tls-ca-bundle.pem /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem
             fi
+            echo -b "Rotating audit logs in /var/log/openshift-apiserver"
```
p0lyn0mial, line 60: what does `-b` do?

bindata/v3.11.0/openshift-apiserver/deploy.yaml
```diff
@@ -68,6 +57,9 @@ spec:
...
+            test -f /var/log/openshift-apiserver/audit.log && mv /var/log/openshift-apiserver/audit{,-$(date '+%Y-%m-%dT%H-%M-%S.%3N')}.log
+            echo
```
p0lyn0mial, line 62: should be removed?

bindata/v3.11.0/openshift-apiserver/deploy.yaml
```diff
@@ -68,6 +57,9 @@ spec:
               echo "Copying system trust bundle"
               cp -f /var/run/configmaps/trusted-ca-bundle/tls-ca-bundle.pem /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem
             fi
+            echo -b "Rotating audit logs in /var/log/openshift-apiserver"
+            test -f /var/log/openshift-apiserver/audit.log && mv /var/log/openshift-apiserver/audit{,-$(date '+%Y-%m-%dT%H-%M-%S.%3N')}.log
```
p0lyn0mial, line 61: how will it work in a single node env?

pod1 writes to  `/var/log/openshift-apiserver/audit.log`
pod2 comes along and renames it to `/var/log/openshift-apiserver/audit-XYZ.log`

what will `pod1` do if there is no file?
what will `pod1` do when `pod2` creates the file?

PR #455, opened 2021-06-09
Bug 1879182: audit policies now use 'secure-' by default, remove custom getter 

stlaz:
Use the "normal" audit policies, the "secure-" type was moved into those.

/assign @sttts


stlaz:
/retest
unrelated

PR #459, opened 2021-06-30
audit: wire auditpolicy controller

sttts:


pkg/operator/starter.go
```diff
@@ -235,6 +234,13 @@ func RunOperator(ctx context.Context, controllerConfig *controllercmd.Controller
 		kubeClient.CoreV1(),
 		kubeClient.CoreV1(),
 		kubeInformersForNamespaces,
+	).WithAuditPolicyController(
+		operatorclient.TargetNamespace,
+		"audit",
```
p0lyn0mial, line 239: ok, this configmap will be mounted here https://github.com/openshift/cluster-openshift-apiserver-operator/blob/master/bindata/v3.11.0/openshift-apiserver/deploy.yaml#L167

PR #465, opened 2021-08-20
Bug 1960680: bindata: run openshift-apiserver as root explicitly.

s-urbaniak:
openshift-apiserver needs to run as root in order to be able to write audit to /var/log/openshift-apiserver.
Currently only "priviledged: true" is specified which can lead to picking SCCs that do not have the RunAsAny policy.

This fixes it by specifying runAsUser: 0 explicitly.

/cc @stlaz @sttts


slaskawi (approved):
/lgtm

PR #466, opened 2021-08-20
Bug 1959290: bump to library-go and kube libraries

sttts:


go.mod
```diff
@@ -7,29 +7,22 @@ require (
...
+	k8s.io/client-go v0.22.1
+	k8s.io/component-base v0.22.1
+	k8s.io/klog/v2 v2.9.0
+	k8s.io/kube-aggregator v0.22.1
+	k8s.io/utils v0.0.0-20210707171843-4b05e18ac7d9
 	sigs.k8s.io/kube-storage-version-migrator v0.0.4
 )
-
-replace (
-	github.com/coreos/etcd => github.com/coreos/etcd v3.3.17+incompatible // Pinned for openshift
```
aojea, line 32: no longer needs this to be pinned?

sttts: see thread in group slack, have asked @hexfusion

hexfusion: this is cruft the test should be based on 3,5 client.

PR #489, opened 2022-01-12
bug 2039670: Create the PDB dynamically so it does not get created on the SNO topology

ingvagabund:
Builds on top of https://github.com/openshift/library-go/pull/1246 and WithIsSNOCheck from https://github.com/openshift/library-go/pull/1238.

### DIFF
--- a/pkg/operator/starter.go
+++ b/pkg/operator/starter.go
@@ -20,6 +20,7 @@ import (
 	"github.com/openshift/library-go/pkg/operator/events"
 	"github.com/openshift/library-go/pkg/operator/genericoperatorclient"
 	"github.com/openshift/library-go/pkg/operator/staleconditions"
+	"github.com/openshift/library-go/pkg/operator/staticpod/controller/guard"
 	"github.com/openshift/library-go/pkg/operator/staticpod/controller/revision"
 	"github.com/openshift/library-go/pkg/operator/status"
 	"github.com/openshift/library-go/pkg/operator/v1helpers"
@@ -210,7 +211,35 @@ func RunOperator(ctx context.Context, controllerConfig *controllercmd.Controller
 					"v3.11.0/openshift-apiserver/svc.yaml",
 					"v3.11.0/openshift-apiserver/sa.yaml",
 					"v3.11.0/openshift-apiserver/trusted_ca_cm.yaml",
+				},
+			},
+			{
+				Files: []string{
 					"v3.11.0/openshift-apiserver/pdb.yaml",
+				},
+				ShouldCreateFn: func() bool {
+					isSNO, precheckSucceeded, err := guard.IsSNOCheckFnc(configInformers.Config().V1().Infrastructures())()
+					if err != nil {
+						klog.Errorf("IsSNOCheckFnc failed: %v", err)
+						return false
+					}
+					if !precheckSucceeded {
+						klog.V(4).Infof("IsSNOCheckFnc precheck did not succeed, skipping")
+						return false
+					}
+					return !isSNO
+				},
+				ShouldDeleteFn: func() bool {
+					isSNO, precheckSucceeded, err := guard.IsSNOCheckFnc(configInformers.Config().V1().Infrastructures())()
+					if err != nil {
+						klog.Errorf("IsSNOCheckFnc failed: %v", err)
+						return false
+					}
+					if !precheckSucceeded {
+						klog.V(4).Infof("IsSNOCheckFnc precheck did not succeed, skipping")
+						return false
+					}
+					return isSNO
 				},
 			},
 		},
--- a/go.mod
+++ b/go.mod
@@ -9,7 +9,7 @@ require (
 	github.com/openshift/api v0.0.0-20211209135129-c58d9f695577
 	github.com/openshift/build-machinery-go v0.0.0-20211213093930-7e33a7eb4ce3
 	github.com/openshift/client-go v0.0.0-20211209144617-7385dd6338e3
-	github.com/openshift/library-go v0.0.0-20220117173518-ca57b619b5d6
+	github.com/openshift/library-go v0.0.0-20220119122101-7e1811fdb665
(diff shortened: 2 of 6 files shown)


pkg/operator/starter.go
```diff
@@ -152,6 +153,8 @@ func RunOperator(ctx context.Context, controllerConfig *controllercmd.Controller
 		statusControllerOptions = append(statusControllerOptions, apiservercontrollerset.WithStatusControllerPdbCompatibleHighInertia("APIServer"))
 	}
 
+	infraLister := configInformers.Config().V1().Infrastructures().Lister()
```
p0lyn0mial, line 156: unused?

ingvagabund: Used in `guard.IsSNOCheckFnc(infraLister)()` on lines 223 and 230

p0lyn0mial: ah, sorry, it is used on line 223 and 230

pkg/operator/starter.go
```diff
@@ -210,8 +213,26 @@ func RunOperator(ctx context.Context, controllerConfig *controllercmd.Controller
...
 					"v3.11.0/openshift-apiserver/sa.yaml",
 					"v3.11.0/openshift-apiserver/trusted_ca_cm.yaml",
+				},
+			},
+			{
+				Files: []string{
 					"v3.11.0/openshift-apiserver/pdb.yaml",
 				},
+				ShouldCreateFn: func() bool {
+					isSno, err := guard.IsSNOCheckFnc(infraLister)()
```
p0lyn0mial, line 223: okay, so it will conditionally create the pdb.yaml based on the current plantform

pkg/operator/starter.go
p0lyn0mial: do we have to wait for `infraLister` to be initialized?

pkg/operator/starter.go
```diff
@@ -210,8 +213,26 @@ func RunOperator(ctx context.Context, controllerConfig *controllercmd.Controller
...
+					if err != nil {
```
p0lyn0mial, line 224: would it make sense to log the error?

pkg/operator/starter.go
ingvagabund: Initialized at https://github.com/openshift/cluster-openshift-apiserver-operator/blob/master/pkg/operator/starter.go#L319

pkg/operator/starter.go
```diff
@@ -321,6 +342,8 @@ func RunOperator(ctx context.Context, controllerConfig *controllercmd.Controller
 	migrationInformer.Start(ctx.Done())
 	apiextensionsInformers.Start(ctx.Done())
 
+	configInformers.WaitForCacheSync(ctx.Done())
```
p0lyn0mial, line 345: oh, it will block all the other controllers. How about extending the `IsSNOCheckFnc` function to accept `HasSynced` instead?

p0lyn0mial: we could simply return `false` before the informer has synced.

p0lyn0mial: and then maybe we could log "real" errors in https://github.com/openshift/cluster-openshift-apiserver-operator/pull/489/files#r782969351

p0lyn0mial: wdyt ?

pkg/operator/starter.go
ingvagabund: Adding klog.Errorf

pkg/operator/starter.go
ingvagabund: Extending the create/delete functions with `infraInformer.HasSynced()` check.

pkg/operator/starter.go
```diff
@@ -210,8 +214,34 @@ func RunOperator(ctx context.Context, controllerConfig *controllercmd.Controller
...
 					"v3.11.0/openshift-apiserver/sa.yaml",
 					"v3.11.0/openshift-apiserver/trusted_ca_cm.yaml",
+				},
+			},
+			{
+				Files: []string{
 					"v3.11.0/openshift-apiserver/pdb.yaml",
 				},
+				ShouldCreateFn: func() bool {
+					if !infraInformer.HasSynced() {
```
p0lyn0mial, line 224: I'd replace with `configInformers.Config().V1().Infrastructures().Informer().HasSynced()`. No need to intro a new variable.

pkg/operator/starter.go
```diff
@@ -210,8 +214,34 @@ func RunOperator(ctx context.Context, controllerConfig *controllercmd.Controller
...
+						klog.Errorf("IsSNOCheck failed: %v", err)
+						return false
+					}
+					return !isSno
+				},
+				ShouldDeleteFn: func() bool {
+					if !infraInformer.HasSynced() {
+						return false
+					}
+					isSno, err := guard.IsSNOCheckFnc(infraLister)()
```
p0lyn0mial, line 238: same here, I'd just `configInformers.Config().V1().Infrastructures().Lister()`

pkg/operator/starter.go
p0lyn0mial: please remove this line

pkg/operator/starter.go
```diff
@@ -152,6 +153,9 @@ func RunOperator(ctx context.Context, controllerConfig *controllercmd.Controller
 		statusControllerOptions = append(statusControllerOptions, apiservercontrollerset.WithStatusControllerPdbCompatibleHighInertia("APIServer"))
 	}
 
+	infraLister := configInformers.Config().V1().Infrastructures().Lister()
+	infraInformer := configInformers.Config().V1().Infrastructures().Informer()
```
p0lyn0mial, line 157: it could be removed if we do https://github.com/openshift/cluster-openshift-apiserver-operator/pull/489/files#r783729064

pkg/operator/starter.go
ingvagabund: Done

PR #494, opened 2022-03-25
Bug 2066886: make apiserver readiness probe honor readyz

deads2k:
This also uses a startup probe instead of a long initial delay seconds for livez and instead of excessive failure thresholds for readiness.

/assign @tkashem

### DIFF
--- a/bindata/v3.11.0/openshift-apiserver/deploy.yaml
+++ b/bindata/v3.11.0/openshift-apiserver/deploy.yaml
@@ -115,17 +115,35 @@ spec:
         - mountPath: /var/log/openshift-apiserver
           name: audit-dir
         livenessProbe:
-          initialDelaySeconds: 30
           httpGet:
             scheme: HTTPS
             port: 8443
             path: healthz
+          initialDelaySeconds: 0
+          periodSeconds: 10
+          timeoutSeconds: 1
+          successThreshold: 1
+          failureThreshold: 3
         readinessProbe:
-          failureThreshold: 10
+          httpGet:
+            scheme: HTTPS
+            port: 8443
+            path: readyz
+          initialDelaySeconds: 0
+          periodSeconds: 5
+          timeoutSeconds: 1
+          successThreshold: 1
+          failureThreshold: 1
+        startupProbe:
           httpGet:
             scheme: HTTPS
             port: 8443
             path: healthz
+          initialDelaySeconds: 0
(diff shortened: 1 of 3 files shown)


bindata/v3.11.0/openshift-apiserver/deploy.yaml
```diff
@@ -115,17 +115,23 @@ spec:
...
+            scheme: HTTPS
+            port: 8443
+            path: readyz
+          periodSeconds: 5
+        startupProbe:
           httpGet:
             scheme: HTTPS
             port: 8443
             path: healthz
+          failureThreshold: 150
```
tkashem, line 133: `1500s`  before the startup probe fails, when does the startup probe start ticking, when the container process starts running?

bindata/v3.11.0/openshift-apiserver/deploy.yaml
```diff
@@ -115,17 +115,23 @@ spec:
         - mountPath: /var/log/openshift-apiserver
           name: audit-dir
         livenessProbe:
-          initialDelaySeconds: 30
           httpGet:
             scheme: HTTPS
             port: 8443
             path: healthz
         readinessProbe:
-          failureThreshold: 10
```
tkashem, line 124: the default is `3`, so it will take `15s` to deem it unready. 

we also have these related settings:
```
  shutdown-delay-duration:
  - 10s # give SDN some time to converge
  shutdown-send-retry-after:
  - "true"
```
https://github.com/openshift/cluster-openshift-apiserver-operator/blob/master/bindata/v3.11.0/config/defaultconfig.yaml#L17-L20

then we should change `shutdown-delay-duration` to `20s` maybe? 15s for pod ready to be fale and then endpoint controller to remove the pod IP

tkashem (commented):
nit: can we also list the fields that we are leaving as defaults? it's easier to do the timeout calculation easily then. 
default values:

```
initialDelaySeconds: 0s
periodSeconds: 10s
timeoutSeconds: 1s
successThreshold: 1
failureThreshold: 3
```
source: https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/

tkashem:
/retitle Bug 2066886: make apiserver readiness probe honor readyz

tkashem:
/lgtm

Thanks David!

PR #511, opened 2023-01-03
OCPBUGS-5300: routes/status resources can leak sensitive data, exclude it from audit

tkashem:
re-vendor  library-go


dgrisonnet (approved):
/lgtm

PR #513, opened 2023-01-04
OCPBUGS-3290: routes/status resources can leak sensitive data, exclude it from audit

tkashem:
re-vendor library-go


go.mod
```diff
@@ -6,22 +6,22 @@ require (
...
+	go.etcd.io/etcd/client/v3 v3.5.1
+	k8s.io/api v0.24.0
+	k8s.io/apiextensions-apiserver v0.24.0
+	k8s.io/apimachinery v0.24.0
+	k8s.io/apiserver v0.24.0
+	k8s.io/client-go v0.24.0
+	k8s.io/component-base v0.24.0
+	k8s.io/klog/v2 v2.60.1
+	k8s.io/kube-aggregator v0.24.0
+	k8s.io/utils v0.0.0-20220210201930-3a6ce19ff2f9
```
dgrisonnet, line 24: I don't think we want to update to the 0.24.0 clients since this is the version of k8s tied to OCP 4.12

go.mod
```diff
@@ -6,22 +6,22 @@ require (
 	github.com/ghodss/yaml v1.0.0
 	github.com/go-bindata/go-bindata v3.1.2+incompatible
 	github.com/gonum/graph v0.0.0-20170401004347-50b27dea7ebb
-	github.com/openshift/api v0.0.0-20211209135129-c58d9f695577
+	github.com/openshift/api v0.0.0-20220525145417-ee5b62754c68
```
dgrisonnet, line 9: do we really need to update the api?

go.mod
```diff
@@ -6,22 +6,22 @@ require (
...
 	github.com/openshift/build-machinery-go v0.0.0-20211213093930-7e33a7eb4ce3
-	github.com/openshift/client-go v0.0.0-20211209144617-7385dd6338e3
-	github.com/openshift/library-go v0.0.0-20220119122101-7e1811fdb665
-	github.com/spf13/cobra v1.2.1
+	github.com/openshift/client-go v0.0.0-20220525160904-9e1acff93e4a
```
dgrisonnet, line 11: do we really need to update the client?

go.mod
tkashem: I think it came through indirect dependencies. https://github.com/openshift/library-go/blob/release-4.11/go.mod points to `v0.24.0`
4.12 is ` v0.25.0`

dgrisonnet: ah correct

PR #532, opened 2023-05-02
WRKLDS-728: Capabilities: drop build/apps APIService when capabilities are not enabled

ingvagabund:
Unless a capability is unknown, install APIService object for corresponding API only when enabled.


pkg/operator/workload/workload_openshiftapiserver_v311_00_sync.go
```diff
@@ -273,15 +279,45 @@ func manageOpenShiftAPIServerImageImportCA_v311_00_to_latest(ctx context.Context
...
+
+	knownCaps := sets.New[configv1.ClusterVersionCapability](clusterVersion.Status.Capabilities.KnownCapabilities...)
+	capsEnabled := sets.New[configv1.ClusterVersionCapability](clusterVersion.Status.Capabilities.EnabledCapabilities...)
+
+	apiServersConfig := openshiftcontrolplanev1.OpenShiftAPIServerConfig{
+		APIServers: openshiftcontrolplanev1.APIServers{
+			PerGroupOptions: []openshiftcontrolplanev1.PerGroupOptions{},
+		},
+	}
+	if knownCaps.Has(configv1.ClusterVersionCapabilityBuild) && !capsEnabled.Has(configv1.ClusterVersionCapabilityBuild) {
```
p0lyn0mial, line 299: Will you have a test that will test the disabling of this APIs?

pkg/operator/workload/workload_openshiftapiserver_v311_00_sync.go
```diff
@@ -273,15 +279,45 @@ func manageOpenShiftAPIServerImageImportCA_v311_00_to_latest(ctx context.Context
...
+		klog.Infof("Capability %q not enabled, disabling 'openshift.io/build' controller", configv1.ClusterVersionCapabilityBuild)
+		apiServersConfig.APIServers.PerGroupOptions = append(apiServersConfig.APIServers.PerGroupOptions, openshiftcontrolplanev1.PerGroupOptions{Name: openshiftcontrolplanev1.OpenShiftBuildAPIserver, DisabledVersions: []string{"v1"}})
+	}
+
+	if knownCaps.Has(configv1.ClusterVersionCapabilityDeploymentConfig) && !capsEnabled.Has(configv1.ClusterVersionCapabilityDeploymentConfig) {
+		klog.Infof("Capability %q not enabled, disabling 'openshift.io/apps' controller", configv1.ClusterVersionCapabilityDeploymentConfig)
```
p0lyn0mial, line 305: won't this spam logs on every `Sync` ?

pkg/operator/workload/workload_openshiftapiserver_v311_00_sync.go
```diff
@@ -273,15 +279,45 @@ func manageOpenShiftAPIServerImageImportCA_v311_00_to_latest(ctx context.Context
 	return resourceapply.ApplyConfigMap(ctx, client, recorder, requiredConfigMap)
 }
 
-func manageOpenShiftAPIServerConfigMap_v311_00_to_latest(ctx context.Context, client coreclientv1.ConfigMapsGetter, recorder events.Recorder, operatorConfig *operatorv1.OpenShiftAPIServer) (*corev1.ConfigMap, bool, error) {
+func manageOpenShiftAPIServerConfigMap_v311_00_to_latest(ctx context.Context, client coreclientv1.ConfigMapsGetter, clusterVersionInformer configinformersconfigv1.ClusterVersionInformer, recorder events.Recorder, operatorConfig *operatorv1.OpenShiftAPIServer) (*corev1.ConfigMap, bool, error) {
 	configMap := resourceread.ReadConfigMapV1OrDie(v311_00_assets.MustAsset("v3.11.0/openshift-apiserver/cm.yaml"))
 	defaultConfig := v311_00_assets.MustAsset("v3.11.0/config/defaultconfig.yaml")
+
+	clusterVersion, err := clusterVersionInformer.Lister().Get("version")
```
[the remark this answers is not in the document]
p0lyn0mial: And whether we need to add a reactor to the controller.

ingvagabund:
> [sig-cli] oc adm must-gather runs successfully for audit logs [apigroup:config.openshift.io][apigroup:oauth.openshift.io] [Suite:openshift/conformance/parallel]

The last two e2e-aws-own failed due to empty openshift-apiserver audit logs

pkg/operator/workload/workload_openshiftapiserver_v311_00_sync.go
ingvagabund: Already in https://github.com/openshift/cluster-openshift-apiserver-operator/pull/532/commits/dc8a55f7c7352a5d0f1e8599a042f13aa913cab6#diff-25e7c9a2521d15c93e122ab635c1ec19358116650aa275abca5e014b514c69c2R241

pkg/operator/workload/workload_openshiftapiserver_v311_00_sync.go
```diff
@@ -273,15 +279,45 @@ func manageOpenShiftAPIServerImageImportCA_v311_00_to_latest(ctx context.Context
...
+		return nil, false, fmt.Errorf("unable to marshal OpenShiftAPIServerConfig struct: %v", err)
+	}
+
 	requiredConfigMap, _, err := resourcemerge.MergePrunedConfigMap(
 		&openshiftcontrolplanev1.OpenShiftAPIServerConfig{},
 		configMap,
 		"config.yaml",
 		nil,
 		defaultConfig,
+		bytes,
```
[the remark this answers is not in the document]
ingvagabund: https://github.com/openshift/cluster-openshift-apiserver-operator/pull/532/commits/dc8a55f7c7352a5d0f1e8599a042f13aa913cab6#diff-25e7c9a2521d15c93e122ab635c1ec19358116650aa275abca5e014b514c69c2R381-R384

pkg/operator/starter.go
```diff
@@ -383,21 +400,26 @@ func RunOperator(ctx context.Context, controllerConfig *controllercmd.Controller
...
-		{Group: "authorization.openshift.io", Version: "v1"},
-		{Group: "build.openshift.io", Version: "v1"},
-		{Group: "image.openshift.io", Version: "v1"},
-		{Group: "project.openshift.io", Version: "v1"},
-		{Group: "quota.openshift.io", Version: "v1"},
-		{Group: "route.openshift.io", Version: "v1"},
-		{Group: "security.openshift.io", Version: "v1"},
-		{Group: "template.openshift.io", Version: "v1"},
+func apiServices(clusterVersionInformer configinformersconfigv1.ClusterVersionInformer) ([]*apiregistrationv1.APIService, []*apiregistrationv1.APIService, error) {
+	clusterVersion, err := clusterVersionInformer.Lister().Get("version")
```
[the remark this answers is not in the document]
p0lyn0mial: I think that we should add `clusterVersionInformer` to `WithInformers` here https://github.com/openshift/library-go/blob/75c7d51fc4155264ba71551d22d0a1968b7bf989/pkg/operator/apiserver/controller/apiservice/apiservice_controller.go#L66 because 

```
// WithInformers is used to register event handlers and get the caches synchronized functions.
```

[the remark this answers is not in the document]
p0lyn0mial: and maybe we should change `GetAPIServicesToMangeFunc` to accept `clusterVersionInformer configinformersconfigv1.ClusterVersionInformer` ?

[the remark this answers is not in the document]
p0lyn0mial: alternative would be to add `WaitForCacheSync` to the `apiServices` function and somehow trigger the controller to call `apiServices` when `clusterVersionInformer` changes/get an event.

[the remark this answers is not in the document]
p0lyn0mial: does it make sense?

pkg/operator/workload/workload_openshiftapiserver_v311_00_sync.go
p0lyn0mial: ping

pkg/operator/starter.go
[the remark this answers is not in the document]
ingvagabund: > I think that we should add clusterVersionInformer to WithInformers here

https://github.com/openshift/library-go/pull/1562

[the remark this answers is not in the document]
ingvagabund: +1 for passing informers as the list of managed api services may have many different implementation which do not need clusterVersionInformer.

pkg/operator/starter.go
```diff
@@ -418,16 +440,20 @@ func apiServices() []*apiregistrationv1.APIService {
...
 
-	return ret
+	return enabled, disabled, nil
 }
 
 func apiServicesReferences() []configv1.ObjectReference {
 	ret := []configv1.ObjectReference{}
-	for _, apiService := range apiServices() {
-		ret = append(ret, configv1.ObjectReference{Group: "apiregistration.k8s.io", Resource: "apiservices", Name: apiService.Spec.Version + "." + apiService.Spec.Group})
+	for _, apiService := range apiServiceGroupVersions {
```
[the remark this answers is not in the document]
ingvagabund: `apiServicesReferences()` is consumed by `.WithClusterOperatorStatusController` which uses the references to build a mapping of objects. In case a reference has no related object, it gets ignored.

pkg/operator/workload/workload_openshiftapiserver_v311_00_sync.go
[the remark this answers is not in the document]
ingvagabund: That's where informer passed in https://github.com/openshift/cluster-openshift-apiserver-operator/pull/532/commits/dc8a55f7c7352a5d0f1e8599a042f13aa913cab6#diff-0d623dfd885adb20f991bda4c2453aebd732ca6dbb4d1d4be6e79805c3b48de6R243 takes affect. Everytime a cluster version changes, OpenShiftAPIServerWorkload.Sync triggers.

pkg/operator/workload/workload_openshiftapiserver_v311_00_sync.go
ingvagabund: I will increase the log level to 4

pkg/operator/starter.go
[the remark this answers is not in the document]
p0lyn0mial: ok, it looks like this is purely informational. It doesn't have any impact on the core logic.

pkg/operator/starter.go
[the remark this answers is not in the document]
p0lyn0mial: ok, thanks.

pkg/operator/workload/workload_openshiftapiserver_v311_00_sync.go
p0lyn0mial: I think that we should have an e2e test for it, wdyt?

pkg/operator/workload/workload_openshiftapiserver_v311_00_sync.go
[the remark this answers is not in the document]
p0lyn0mial: okay, so it looks like we don't need an informer, we need a Lister here :) Please change it to a Lister.

pkg/operator/workload/workload_openshiftapiserver_v311_00_sync_test.go
```diff
@@ -94,20 +99,33 @@ func TestOperatorConfigProgressingCondition(t *testing.T) {
...
 				countNodes:                fakeCountNodes,
 				ensureAtMostOnePodPerNode: func(spec *appsv1.DeploymentSpec, componentName string) error { return nil },
 			}
 
-			if _, _, err := target.Sync(context.Background(), factory.NewSyncContext("TestSyncCOntext", events.NewInMemoryRecorder(""))); len(err) > 0 {
+			// register the informer
+			configInformers.Config().V1().ClusterVersions().Informer()
+
+			ctx := context.Background()
+			configInformers.Start(ctx.Done())
```
p0lyn0mial, line 125: and having a lister instead of an informer will simplify this unit tests :)

pkg/operator/workload/workload_openshiftapiserver_v311_00_sync.go
```diff
@@ -273,15 +279,46 @@ func manageOpenShiftAPIServerImageImportCA_v311_00_to_latest(ctx context.Context
...
+		klog.V(4).Infof("Capability %q not enabled, disabling 'openshift.io/build' controller", configv1.ClusterVersionCapabilityBuild)
+		apiServersConfig.APIServers.PerGroupOptions = append(apiServersConfig.APIServers.PerGroupOptions, openshiftcontrolplanev1.PerGroupOptions{Name: openshiftcontrolplanev1.OpenShiftBuildAPIserver, DisabledVersions: []string{"v1"}})
+	}
+
+	if knownCaps.Has(configv1.ClusterVersionCapabilityDeploymentConfig) && !capsEnabled.Has(configv1.ClusterVersionCapabilityDeploymentConfig) {
+		klog.V(4).Infof("Capability %q not enabled, disabling 'openshift.io/apps' controller", configv1.ClusterVersionCapabilityDeploymentConfig)
+		apiServersConfig.APIServers.PerGroupOptions = append(apiServersConfig.APIServers.PerGroupOptions, openshiftcontrolplanev1.PerGroupOptions{Name: openshiftcontrolplanev1.OpenShiftAppsAPIserver, DisabledVersions: []string{"v1"}})
+	}
+
+	bytes, err := json.Marshal(apiServersConfig)
```
p0lyn0mial, line 310: mhm, should we be passing a yaml ?

pkg/operator/workload/workload_openshiftapiserver_v311_00_sync_test.go
```diff
@@ -219,3 +237,155 @@ func TestPreconditionFulfilled(t *testing.T) {
...
+			configInformers.WaitForCacheSync(ctx.Done())
+
+			if _, _, err := target.Sync(ctx, factory.NewSyncContext("TestSyncCOntext", events.NewInMemoryRecorder(""))); len(err) > 0 {
+				t.Fatal(err)
+			}
+
+			cm, err := kubeClient.CoreV1().ConfigMaps("openshift-apiserver").Get(ctx, "config", metav1.GetOptions{})
```
p0lyn0mial, line 376: should we check `kubeClient.Actions()` instead ?

pkg/operator/workload/workload_openshiftapiserver_v311_00_sync_test.go
```diff
@@ -219,3 +237,155 @@ func TestPreconditionFulfilled(t *testing.T) {
...
+				{Name: openshiftcontrolplanev1.OpenShiftBuildAPIserver, DisabledVersions: []string{"v1"}},
+				{Name: openshiftcontrolplanev1.OpenShiftAppsAPIserver, DisabledVersions: []string{"v1"}},
+			},
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+
+			kubeClient := fake.NewSimpleClientset(
```
p0lyn0mial, line 309: please rename it to `fakeKubeClient`

pkg/operator/workload/workload_openshiftapiserver_v311_00_sync_test.go
```diff
@@ -219,3 +237,155 @@ func TestPreconditionFulfilled(t *testing.T) {
...
+				t.Fatal(err)
+			}
+
+			cm, err := kubeClient.CoreV1().ConfigMaps("openshift-apiserver").Get(ctx, "config", metav1.GetOptions{})
+			if err != nil {
+				t.Fatalf("Unable to get 'config' configmap: %v", err)
+			}
+
+			config := openshiftcontrolplanev1.OpenShiftAPIServerConfig{}
+			if err := json.NewDecoder(bytes.NewBuffer([]byte(cm.Data["config.yaml"]))).Decode(&config); err != nil {
```
p0lyn0mial, line 382: `oas` uses a versioned decoder. Could we do the same here ?

```
scheme := runtime.NewScheme()
utilruntime.Must(openshiftcontrolplanev1.Install(scheme))
codecs := serializer.NewCodecFactory(scheme)
obj, err := runtime.Decode(codecs.UniversalDecoder(openshiftcontrolplanev1.GroupVersion, configv1.GroupVersion), configContent)
if err != nil {
	return err
}
config := obj.(*openshiftcontrolplanev1.OpenShiftAPIServerConfig)
```


https://github.com/openshift/openshift-apiserver/blob/df3ca642f426cf2f34abeae152a9fce80b44c63e/pkg/cmd/openshift-apiserver/cmd.go#L129

pkg/operator/workload/workload_openshiftapiserver_v311_00_sync_test.go
```diff
@@ -219,3 +237,155 @@ func TestPreconditionFulfilled(t *testing.T) {
 		})
 	}
 }
+
+func TestCapabilities(t *testing.T) {
+	testCases := []struct {
+		name                    string
+		knownCapabilities       []configv1.ClusterVersionCapability
+		enabledCapabilities     []configv1.ClusterVersionCapability
+		expectedPerGroupOptions []openshiftcontrolplanev1.PerGroupOptions
```
p0lyn0mial, line 246: would it make sense to validate the whole config and not only `PerGroupOptions` ?
assuming that other fields remain unchanged, this shouldn't require much more effort

pkg/operator/workload/workload_openshiftapiserver_v311_00_sync.go
ingvagabund: Updating/adding e2es is the next step once this PR and https://github.com/openshift/cluster-openshift-controller-manager-operator/pull/291 are merged.

pkg/operator/workload/workload_openshiftapiserver_v311_00_sync_test.go
ingvagabund: Done

pkg/operator/workload/workload_openshiftapiserver_v311_00_sync_test.go
ingvagabund: Done

pkg/operator/workload/workload_openshiftapiserver_v311_00_sync.go
p0lyn0mial: where are you planing to add that test ? will it be this repo or in the origin ?

pkg/operator/workload/workload_openshiftapiserver_v311_00_sync_test.go
ingvagabund: What for?

pkg/operator/workload/workload_openshiftapiserver_v311_00_sync_test.go
ingvagabund: TestOperatorConfigProgressingCondition uses kubeClient well. Do you wanna rename it on other test(s) as well?

pkg/operator/workload/workload_openshiftapiserver_v311_00_sync.go
ingvagabund: Origin as we need to check for disabled controllers as well. E.g. read the config or parse the controllers logs.

pkg/operator/workload/workload_openshiftapiserver_v311_00_sync_test.go
p0lyn0mial: We usually do this to check what resource would be sent to the server and how many times. The Get method reads the state stored in memory, which may not necessarily correspond to the resource that would be sent to the server.

pkg/operator/workload/workload_openshiftapiserver_v311_00_sync_test.go
ingvagabund: That would require to construct a full config to compare it with. Also, diff.ObjectDiff over `config.APIServers.PerGroupOptions` is easier to read.

pkg/operator/workload/workload_openshiftapiserver_v311_00_sync_test.go
p0lyn0mial: sure, if this is not a problem for you that would be fine :)

pkg/operator/workload/workload_openshiftapiserver_v311_00_sync.go
p0lyn0mial: Ideally if we could add a test here and then also run it from the origin repo. Are you planing to add the test before we ship 4.14 ?

pkg/operator/workload/workload_openshiftapiserver_v311_00_sync_test.go
ingvagabund: Something like https://github.com/openshift/openshift-apiserver/blob/064c2d0ef0ecaeda2bcc4387eaaa7258cee5adcf/pkg/project/apiserver/registry/project/proxy/proxy_test.go#L96-L101? Is the goal here to add just a test for `Matches("update", "configmaps")`? Or, to more inspect the corresponding action? Do you have an example?

pkg/operator/workload/workload_openshiftapiserver_v311_00_sync.go
ingvagabund: Yes. All tests need to land before 4.14 ships.

pkg/operator/workload/workload_openshiftapiserver_v311_00_sync_test.go
p0lyn0mial: Usually we verify the verbs (https://github.com/openshift/library-go/blob/abaa55e36ec42bcade37586fbeabfb5bf48cd5b8/pkg/operator/encryption/controllers/key_controller_test.go#L389)

And the content of the resource as well (https://github.com/openshift/library-go/blob/abaa55e36ec42bcade37586fbeabfb5bf48cd5b8/pkg/operator/encryption/controllers/key_controller_test.go#L302)

I think we could check both here as well.

pkg/operator/workload/workload_openshiftapiserver_v311_00_sync.go
p0lyn0mial: OK. I would like to avoid shipping a feature without the tests. Thanks.

pkg/operator/workload/workload_openshiftapiserver_v311_00_sync_test.go
ingvagabund: Done

pkg/operator/workload/workload_openshiftapiserver_v311_00_sync_test.go
ingvagabund: Done

pkg/operator/workload/workload_openshiftapiserver_v311_00_sync_test.go
ingvagabund: Done

pkg/operator/workload/workload_openshiftapiserver_v311_00_sync.go
ingvagabund: Done

pkg/operator/workload/workload_openshiftapiserver_v311_00_sync.go
```diff
@@ -273,15 +279,46 @@ func manageOpenShiftAPIServerImageImportCA_v311_00_to_latest(ctx context.Context
...
+		},
+	}
+
+	if knownCaps.Has(configv1.ClusterVersionCapabilityBuild) && !capsEnabled.Has(configv1.ClusterVersionCapabilityBuild) {
+		klog.V(4).Infof("Capability %q not enabled, disabling 'openshift.io/build' controller", configv1.ClusterVersionCapabilityBuild)
+		apiServersConfig.APIServers.PerGroupOptions = append(apiServersConfig.APIServers.PerGroupOptions, openshiftcontrolplanev1.PerGroupOptions{Name: openshiftcontrolplanev1.OpenShiftBuildAPIserver, DisabledVersions: []string{"v1"}})
+	}
+
+	if knownCaps.Has(configv1.ClusterVersionCapabilityDeploymentConfig) && !capsEnabled.Has(configv1.ClusterVersionCapabilityDeploymentConfig) {
+		klog.V(4).Infof("Capability %q not enabled, disabling 'openshift.io/apps' controller", configv1.ClusterVersionCapabilityDeploymentConfig)
```
p0lyn0mial, line 306: mhm, I still think it will be logged on every sync. What is the value of spamming the log file with this information ? Ideally if we could log it just once, right ?

p0lyn0mial: Maybe we need a new condition ? Does the service controller set a condition ?

ingvagabund: With log 4 it is good to know the operator properly reads the capabilities. Otherwise, I'd need to get the CM, decode it, read the list of disabled apiservers and compare it. Too much additional code for just a single log line.

ingvagabund: The API service controller sets the new Degraded condition for API service endpoints until the disabled APIService objects are deleted. The test is reported through the already existing conditions for operands.

PR #536, opened 2023-06-02
OCPBUGS-14010: increase timeout for probes

tkashem:


deads2k:
/lgtm
/hold

holding in case you'd like another set of eyes.

PR #538, opened 2023-06-06
OCPBUGS-2765: Library go bump

ibihim:
# What

Bumping library-go.
Bumping library-go's necessary deps.

# Why

Contains updated audit policy.


_output/golang-versions
```diff
@@ -0,0 +1,3 @@
+1.20
```
dgrisonnet, line 1: We shoudn't push these files generated in _output

ibihim: oh, nice gotcha.

do you know why they are not part of `.gitignore`?

PR #542, opened 2023-07-06
Plumb featuregates to the openshift-apiserver

deads2k:
This will allow us to have techpreview fields for openshift APIs.

### DIFF
--- a/pkg/operator/starter.go
+++ b/pkg/operator/starter.go
@@ -2,6 +2,7 @@ package operator
 
 import (
 	"context"
+	"fmt"
 	"os"
 	"time"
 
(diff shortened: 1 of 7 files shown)


pkg/operator/starter.go
```diff
@@ -97,6 +98,27 @@ func RunOperator(ctx context.Context, controllerConfig *controllercmd.Controller
 		return err
 	}
 
+	desiredVersion := status.VersionForOperatorFromEnv()
+	missingVersion := "0.0.1-snapshot"
+
+	// By default, this will exit(0) the process if the featuregates ever change to a different set of values.
```
p0lyn0mial, line 104: wow, may I know why the process will exit if the featuregates change?

pkg/operator/starter.go
```diff
@@ -97,6 +98,27 @@ func RunOperator(ctx context.Context, controllerConfig *controllercmd.Controller
...
+	featureGateAccessor := featuregates.NewFeatureGateAccess(
+		desiredVersion, missingVersion,
+		configInformers.Config().V1().ClusterVersions(), configInformers.Config().V1().FeatureGates(),
+		controllerConfig.EventRecorder,
+	)
+	go featureGateAccessor.Run(ctx)
+	go configInformers.Start(ctx.Done())
```
p0lyn0mial, line 111: you don't need to start informers in a separate goroutine
`configInformers.Start(ctx.Done())` works

p0lyn0mial: `configInformers` are started again in L367 (https://github.com/openshift/cluster-openshift-apiserver-operator/blob/e8930013a372a80360467c103a8ddfc1512d3d5f/pkg/operator/starter.go#L367)

please fix it.

pkg/operator/starter.go
```diff
@@ -97,6 +98,27 @@ func RunOperator(ctx context.Context, controllerConfig *controllercmd.Controller
...
+
+	select {
```
p0lyn0mial, line 113: why do we need this blocking select statement ?
why waiting for featuregates cannot be dynamic?

p0lyn0mial: for example:
```
if !f.featureGateAccess.AreInitialFeatureGatesObserved() {
		// if we haven't observed featuregates yet, return the existing
		return prunedExistingConfig, nil
	}
```

pkg/operator/starter.go
deads2k: > wow, may I know why the process will exit if the featuregates change?

Simplest way to be sure that every control loop is reconsidered.  And the operators are restartable at any time. And featuregates can only change once in the lifetime of the cluster.

pkg/operator/starter.go
deads2k: The config informers must be started here so this control loop runs.  Doing what you suggest and avoiding a second restart will cause other callers of the shareinformerfactory that start informers against *other* resources to never be started.

pkg/operator/starter.go
deads2k: > why do we need this blocking select statement ?
> why waiting for featuregates cannot be dynamic?

It could be, but it's harder to reason about correctness as control loop specific to techpreview features are enabled. this is the standard block across all oeprators.  Though give the questions above it may be easier for people to reason about if I create a blocking function that takes a rest.Config.

pkg/operator/starter.go
p0lyn0mial: okay, ic, `configInformers` is a factory not an informer :)

PR #544, opened 2023-07-13
AUTH-408: bindata: set required-scc

s-urbaniak:
This sets the required-scc annotation for openshift-apiserver to `privileged` as openshift-apiserver is a privileged pod.

/cc @deads2k @stlaz

### DIFF
--- a/bindata/v3.11.0/openshift-apiserver/deploy.yaml
+++ b/bindata/v3.11.0/openshift-apiserver/deploy.yaml
@@ -40,6 +40,7 @@ spec:
         apiserver: "true"
       annotations:
         target.workload.openshift.io/management: '{"effect": "PreferredDuringScheduling"}'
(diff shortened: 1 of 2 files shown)


s-urbaniak:
Verified that the right SCC is being picked:
```
$ cat namespaces/openshift-apiserver/core/pods.yaml
...
---
apiVersion: v1
items:
- apiVersion: v1
  kind: Pod
  metadata:
    annotations:
...
      openshift.io/required-scc: privileged
      openshift.io/scc: privileged
...
```

s-urbaniak:
just fyi, looking at other PRs it used to be `node-exporter` before :-|

```
---
apiVersion: v1
items:
- apiVersion: v1
  kind: Pod
  metadata:
    annotations:
...
      openshift.io/scc: node-exporter
```

Issue #549, opened 2023-09-27
Return error on RunOperator context done


danielkucera:
When the context is done during RunOperator should not the error returned be not null? Is this not an error?
https://github.com/openshift/cluster-openshift-apiserver-operator/blob/00f7e4cc95063ba5aba1992568088d924cfbf516/pkg/operator/starter.go#L401

danielkucera:
I suggest:
```
	<-ctx.Done()
	return context.Cause(ctx)
}
```

PR #550, opened 2023-09-27
return cause on context done

danielkucera:
see #549


pkg/operator/starter.go
```diff
@@ -398,7 +398,7 @@ func RunOperator(ctx context.Context, controllerConfig *controllercmd.Controller
 	go connectivityCheckController.Run(ctx, 1)
 
 	<-ctx.Done()
-	return nil
+	return context.Cause(ctx)
```
p0lyn0mial, line 401: won't this report an error even when the ctx was cancelled ?
if yes, then do you know when the ctx can be cancelled ?
will the ctx be cancelled when the process receives a `SIGTERM` ?

PR #573, opened 2024-02-20
AUTH-482: Set required-scc for openshift workloads

liouk:


manifests/07_deployment.yaml
```diff
@@ -23,6 +23,7 @@ spec:
         app: openshift-apiserver-operator
       annotations:
         target.workload.openshift.io/management: '{"effect": "PreferredDuringScheduling"}'
+        openshift.io/required-scc: nonroot-v2
```
p0lyn0mial, line 26: how does it differ from `openshift.io/scc` ?

I've checked [a CI run](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_cluster-openshift-apiserver-operator/573/pull-ci-openshift-cluster-openshift-apiserver-operator-master-e2e-aws-operator/1783498460893286400/artifacts/e2e-aws-operator/gather-must-gather/artifacts/) and oas has:
```
  openshift.io/required-scc: privileged
   openshift.io/scc: privileged
 ```
 
 is this expected ?

Comment (p0lyn0mial) [reply manifests/07_deployment.yaml:26]: additionally [e2e-aws-operator](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-openshift-apiserver-operator/573/pull-ci-openshift-cluster-openshift-apiserver-operator-master-e2e-aws-operator/1783498460893286400) reported:

```
{Operator progressing (NodeInstaller): NodeInstallerProgressing: 3 nodes are at revision 9; 0 nodes have achieved new revision 10  Operator progressing (NodeInstaller): NodeInstallerProgressing: 3 nodes are at revision 9; 0 nodes have achieved new revision 10}
```

is this connected to changes in this PR ?

liouk: > how does it differ from `openshift.io/scc` ?
`openshift.io/scc` is the SCC with which the pod is getting admitted; `openshift.io/required-scc` is the SCC with which we request the pod to get admitted with (a.k.a. SCC pinning), so that we basically force the selection of an SCC, if that SCC of course is allowed based on the pod spec.

> is this expected ?
This can happen, yes; I guess the oas has different requirements from the oas-operator. I have chosen `nonroot-v2` because this is the SCC with which the oas-operator pod is getting admitted with.

> is this connected to changes in this PR ?
I doubt this is connected to this PR. Here we're only pinning the SCC that is anyway the one chosen to admit the oas-operator pod, so there shouldn't be any effect. This is just to protect the pod from getting admitted with a custom SCC by accident, should one exist.

p0lyn0mial: >openshift.io/scc is the SCC with which the pod is getting admitted; openshift.io/required-scc is the SCC with which we request the pod to get admitted with (a.k.a. SCC pinning), so that we basically force the selection of an SCC, if that SCC of course is allowed based on the pod spec.

isn't it strange that even though we requested a different scc we got `openshift.io/required-scc: privileged` ? For me it is very confusing because, just by looking at the manifest, I would expect a different annotation to be applied to the final manifest.

p0lyn0mial: ok, we've explained it through slack, the annotation is for `oas-o` and not `oas`. I have one more question about `scc` itself, why `nonroot-v2`? Is this the lowest possible level of permissions?

liouk: > the annotation is for `oas-o` and not `oas`

Correct; the `required-scc` annotation is already defined for `oas` pods.

> Is this the lowest possible level of permissions?

Indeed :+1:

p0lyn0mial: From https://docs.openshift.com/container-platform/4.15/authentication/managing-security-context-constraints.html#default-sccs_configuring-internal-oauth it looks like `nonroot-v2` is better than `nonroot` or just `restricted`

p0lyn0mial:
/lgtm
/approve

thanks!

PR #57, opened 2018-11-20
switch to operator v1

deads2k:
/assign @sanchezl

PR #582, opened 2024-08-16
MULTIARCH-4557: Sync import mode image config status field in the observed config

Prashanth684:
As per https://github.com/openshift/enhancements/pull/1605, there was a
new field introduced in the image config spec and status which reflects
the global value to be set for imagestream import mode which is behind a
featuregate. This PR reads the image config status, checks if the importmode
string is present and syncs that value in the imagepolicyconfig of the
observed config to be used by the apiserver.

API changes: https://github.com/openshift/api/pull/1928
openshift-apiserver changes: https://github.com/openshift/openshift-apiserver/pull/443
image-registry-operator changes: https://github.com/openshift/cluster-image-registry-operator/pull/1090

### DIFF
--- a/pkg/operator/configobservation/images/observe_images.go
+++ b/pkg/operator/configobservation/images/observe_images.go
@@ -148,6 +148,56 @@ func ObserveAllowedRegistriesForImport(genericListers configobserver.Listers, re
 	return observedConfig, errs
 }
 
+func ObserveImagestreamImportMode(genericListers configobserver.Listers, recorder events.Recorder, existingConfig map[string]interface{}) (map[string]interface{}, []error) {
+	listers := genericListers.(configobservation.Listers)
+	var errs []error
+
+	// first observe all the existing config values so that if we get any errors
+	// we can at least return those.
+	imageStreamImportModePath := []string{"imagePolicyConfig", "imageStreamImportMode"}
+	currentImageStreamImportMode, _, err := unstructured.NestedString(existingConfig, imageStreamImportModePath...)
+	if err != nil {
+		return map[string]interface{}{}, append(errs, err)
+	}
+
+	prevObservedConfig := map[string]interface{}{}
+	if len(currentImageStreamImportMode) > 0 {
+		err := unstructured.SetNestedField(prevObservedConfig, currentImageStreamImportMode, imageStreamImportModePath...)
+		if err != nil {
+			return prevObservedConfig, append(errs, err)
+		}
+	}
+
+	// now gather the cluster config and turn it into the observed config
+	observedConfig := map[string]interface{}{}
+	configImage, err := listers.ImageConfigLister.Get("cluster")
+	if errors.IsNotFound(err) {
+		klog.Warningf("ImageStreamImportMode observer: image.config.openshift.io/cluster: not found")
+		return observedConfig, errs
+	}
+	if err != nil {
+		return prevObservedConfig, append(errs, err)
+	}
+
+	imageStreamImportMode := ""
+	if len(string(configImage.Status.ImageStreamImportMode)) > 0 {
+		imageStreamImportMode = string(configImage.Status.ImageStreamImportMode)
+		err = unstructured.SetNestedField(observedConfig, imageStreamImportMode, imageStreamImportModePath...)
+		if err != nil {
+			return prevObservedConfig, append(errs, err)
+		}
+	}
+
+	// no change, return early to skip the event
+	if imageStreamImportMode == currentImageStreamImportMode {
+		return observedConfig, errs
+	}
+
+	recorder.Eventf("ImageStreamImportModeChanged", "ImageStreamImportMode changed from %q to %q", currentImageStreamImportMode, imageStreamImportMode)
+
+	return observedConfig, errs
+}
+
 func Convert(o interface{}) (interface{}, error) {
 	if o == nil {
 		return nil, nil
--- a/pkg/operator/configobservation/images/observe_images_test.go
+++ b/pkg/operator/configobservation/images/observe_images_test.go
@@ -18,10 +18,12 @@ import (
 )
 
 type imageConfigTest struct {
+	name                              string
 	imageConfig                       *configv1.Image
 	expectedInternalRegistryHostname  string
 	expectedExternalRegistryHostnames []string
 	expectedAllowedRegistries         []configv1.RegistryLocation
+	expectedImageStreamImportMode     configv1.ImportModeType
 }
 
 func TestObserveImageConfig(t *testing.T) {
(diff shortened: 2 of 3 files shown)


go.mod
```diff
@@ -6,20 +6,20 @@ require (
...
 	github.com/openshift/library-go v0.0.0-20240412151150-149aa7608754
 	github.com/spf13/cobra v1.7.0
 	github.com/stretchr/testify v1.8.4
 	go.etcd.io/etcd/client/v3 v3.5.10
-	k8s.io/api v0.29.0
+	k8s.io/api v0.30.1
 	k8s.io/apiextensions-apiserver v0.29.0
-	k8s.io/apimachinery v0.29.0
+	k8s.io/apimachinery v0.30.1
 	k8s.io/apiserver v0.29.0
```
dinhxuanvu, line 19: Please also bump other k8s components to make sure they are all on 1.30.1 version.

dinhxuanvu (changes requested):
Overall, looks good to me.

go.mod
Prashanth684: done. had to update an import here: `pkg/operator/revisionpoddeployer/deployers_test.go`

dinhxuanvu (commented):
/lgtm

dinhxuanvu:
@tkashem @p0lyn0mial For final revie and approval.

pkg/operator/configobservation/images/observe_images.go
```diff
@@ -148,6 +148,46 @@ func ObserveAllowedRegistriesForImport(genericListers configobserver.Listers, re
 	return observedConfig, errs
 }
 
+func ObserveImagestreamImportMode(genericListers configobserver.Listers, recorder events.Recorder, existingConfig map[string]interface{}) (map[string]interface{}, []error) {
+	listers := genericListers.(configobservation.Listers)
```
tkashem, line 152: is it possible to use `lister := configobserver.ImageConfigLister` directly?

pkg/operator/configobservation/images/observe_images.go
```diff
@@ -148,6 +148,46 @@ func ObserveAllowedRegistriesForImport(genericListers configobserver.Listers, re
...
+	var errs []error
+	prevObservedConfig := map[string]interface{}{}
+
+	// first observe all the existing config values so that if we get any errors
+	// we can at least return those.
+	imageStreamImportModePath := []string{"imagePolicyConfig", "imageStreamImportMode"}
+	currentImageStreamImportMode, _, err := unstructured.NestedString(existingConfig, imageStreamImportModePath...)
+	if err != nil {
+		return prevObservedConfig, append(errs, err)
```
tkashem, line 161: nit: returning between `prevObservedConfig` and  `existingConfig` is a little confusing to the reader, we are basically saying here `return map[string]interface{}{}, append(errs, err)`

pkg/operator/configobservation/images/observe_images.go
```diff
@@ -148,6 +148,46 @@ func ObserveAllowedRegistriesForImport(genericListers configobserver.Listers, re
...
+		if err != nil {
+			return prevObservedConfig, append(errs, err)
+		}
+	}
+
+	// now gather the cluster config and turn it into the observed config
+	observedConfig := map[string]interface{}{}
+	configImage, err := listers.ImageConfigLister.Get("cluster")
+	if errors.IsNotFound(err) {
+		klog.Warningf("image.config.openshift.io/cluster: not found")
```
tkashem, line 174: can we include more contextual information here, like the observer name?

pkg/operator/configobservation/images/observe_images.go
```diff
@@ -148,6 +148,46 @@ func ObserveAllowedRegistriesForImport(genericListers configobserver.Listers, re
...
+		return observedConfig, errs
```
tkashem, line 175: can you please, perhaps with a comment, clarify why we have to special case `NotFound` here, and return `observedConfig` (empty) instead of `prevObservedConfig`?

pkg/operator/configobservation/images/observe_images.go
```diff
@@ -148,6 +148,46 @@ func ObserveAllowedRegistriesForImport(genericListers configobserver.Listers, re
...
+	}
+	if err != nil {
+		return prevObservedConfig, append(errs, err)
+	}
+
+	imageStreamImportMode := string(configImage.Status.ImageStreamImportMode)
+	if len(imageStreamImportMode) > 0 {
```
tkashem, line 182: if it has changed then we want to add an event, please see: https://github.com/openshift/cluster-openshift-apiserver-operator/blob/ae690fe574be44fae506bfcc59caa629fce9a803/pkg/operator/configobservation/project/observe_projects.go#L64-L69

pkg/operator/configobservation/images/observe_images_test.go
```diff
@@ -121,6 +122,34 @@ func TestObserveImageConfig(t *testing.T) {
 			},
 			expectedAllowedRegistries: nil,
 		},
+		{
```
tkashem, line 125: Can this be a new test? also, I would like to see a description of the use case being tested here, so please consider adding `name` to `imageConfigTest`

tkashem (commented):
Have a few minor comments
/approve
Thanks

pkg/operator/configobservation/images/observe_images.go
Prashanth684: the ImageConfigLister is under the Listers object [here](https://github.com/openshift/cluster-openshift-apiserver-operator/blob/ae690fe574be44fae506bfcc59caa629fce9a803/pkg/operator/configobservation/interfaces.go#L15) which is passed through [here](https://github.com/openshift/cluster-openshift-apiserver-operator/blob/ae690fe574be44fae506bfcc59caa629fce9a803/pkg/operator/configobservation/configobservercontroller/observe_config_controller.go#L39). I believe that is why it is done this way?

pkg/operator/configobservation/images/observe_images.go
Prashanth684: honestly, i followed the pattern of the other observer functions in this file. Going back, it looks like https://github.com/openshift/cluster-openshift-apiserver-operator/pull/57 introduced it. That is a good question but I don't know what the motivation for that was.

Prashanth684: maybe because if the object is not found, it has been potentially removed and in that case we want it to be empty?

tkashem: probably yes, @sanchezl if the `cluster` object (the thing we are observing for config) is missing, is it an error, or the operator does its best to continue with the old configuration?

pkg/operator/configobservation/images/observe_images.go
Prashanth684: done.

pkg/operator/configobservation/images/observe_images.go
Prashanth684: added.

pkg/operator/configobservation/images/observe_images.go
Prashanth684: done.

pkg/operator/configobservation/images/observe_images_test.go
Prashanth684: done.

pkg/operator/configobservation/images/observe_images.go
Prashanth684: @tkashem is my observation correct, or is there another way to do it?

tkashem:
/lgtm

/hold
(cancel if you are not waiting for others to review)

PR #588, opened 2024-09-10
WIP: OCPBUGS-41617: increase openshift-apiserver failureThreshold

liouk:
This PR improves the liveness/readiness checks of the openshift-apiserver.

In particular:
- replace liveness probe path `healthz` with `livez`, and exclude `etcd` from the probe
- increase the readiness probe failure threshold to 4


bindata/v3.11.0/openshift-apiserver/deploy.yaml
```diff
@@ -134,12 +134,12 @@ spec:
           periodSeconds: 5
           timeoutSeconds: 10
```
benluddy, line 135: I know you're not touching this, but I am curious about the implications on graceful shutdown of having a probe timeout that is greater than the period. If every probe request times out, does the period effectively become 10s?

p0lyn0mial: That is a very good question. In the other thread you are suggesting to set `--shutdown-delay-duration` to match the `readinessProbe`. The value for that parameter might be different depending on the proble implementation. Am I right ?

Assuming the period of 5s will be ignored after reaching timeout we should set `--shutdown-delay-duration` to `40s`, am I right?

p0lyn0mial: At the moment the shutdown delay is set to `15s`. Increasing this value will delay rollout significantly.

p0lyn0mial:
would it make sense to set the other parameters in this PR to make backporting easier?

bindata/v3.11.0/openshift-apiserver/deploy.yaml
benluddy: Yes, I think so. The selection of 10s for timeoutSeconds appears to have been based on an etcd check (https://issues.redhat.com//browse/OCPBUGS-14010). Do we need to revisit that choice as part of this? Can we make the etcd check asynchronous so that it isn't on the critical path of the apiserver livez and readyz handlers?

bindata/v3.11.0/openshift-apiserver/deploy.yaml
```diff
@@ -134,12 +134,12 @@ spec:
...
           successThreshold: 1
-          failureThreshold: 1
+          failureThreshold: 4
```
p0lyn0mial, line 137: @liouk, @benluddy left me this message:

```
so, the worst-case assumption is that the apiserver container gets sigterm from kubelet the instant after a successful readiness probe response
so the first post-shutdown-delay readiness probe fires after the probe's interval has elapsed
then kubelet needs to observe <failureThreshold> probe failures before becoming unready
each probe should take between 0 and <timeout> seconds to get a result
i think this means we need to wait <failureThreshold> * (<interval> + <timeout>) to become unready, starting from the time graceful shutdown begins
as the minimum
then we need some buffer on top of that for kubelet to update the pod ready condition, endpoint controller to observe the pod change and reconcile endpoints, and kube-apiserver endpoint informers to observe that change and update the lister caches
```
I think Ben is right. I would lover the `failureThreshold` to `3` and that would give us 3*(5+10) + 15 (time for going unready, removing the endpoint and for kas to observer the change) for shutdown-delay.

bindata/v3.11.0/openshift-apiserver/deploy.yaml
p0lyn0mial: > Do we need to revisit that choice as part of this?

I don't think so, because we would like to set the etcd timeouts to 9s.

> Can we make the etcd check asynchronous so that it isn't on the critical path of the apiserver livez and readyz handlers

we could, but that would require changing upstream first besides we are going to exclude the etcd checks from the livez

liouk: After doing some tests, I've found out that if `timeoutSeconds > periodSeconds`, then the kubelet will issue the next probe right after the timeout (if the previous probe times out, that is), and it won't wait for another `periodSeconds`.

This means that 3 readiness checks will fail in the worst case after `5s + 3*10s`.

bindata/v3.11.0/openshift-apiserver/deploy.yaml
liouk: As explained above, setting it to 3 would give us `5s + 3*10s` so if we add `15s` as a buffer, we get `50s` in total.

p0lyn0mial:
/lgtm

holding so that @benluddy can take a look as well
/hold

p0lyn0mial:
I've checked https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-openshift-apiserver-operator/588/pull-ci-openshift-cluster-openshift-apiserver-operator-master-e2e-aws-operator/1833829309979037696 for `TerminationMinimalShutdownDurationFinished` event and it says `The minimal shutdown duration of 50s finished`

bindata/v3.11.0/openshift-apiserver/deploy.yaml
```diff
@@ -134,12 +134,12 @@ spec:
...
           timeoutSeconds: 10
           successThreshold: 1
-          failureThreshold: 1
+          failureThreshold: 3
         startupProbe:
           httpGet:
             scheme: HTTPS
             port: 8443
-            path: healthz
+            path: livez
```
benluddy, line 142: Shouldn't this follow the liveness probe by excluding the etcd check?

p0lyn0mial: I don't think so. This probe is executed only once and without etcd we cannot start the server.

benluddy: It may not make much difference. Both readyz and livez are gated on post-start hook completion, and we know that many of those require informer syncs. What's the upside to including the etcd check in the startup probe? If etcd is unavailable for some reason, we may as well continue waiting for it to come back instead of failing the startup probe and restarting.

p0lyn0mial: My reasoning is that the startup probe is executed only once and is used to know if the container has successfully started. If the kas  cannot reach etcd during startup, it may indicate a fundamental issue that should block the startup until it is resolved.

PR #592, opened 2024-09-23
API-1835: update for more apply loops

deads2k:
proof of https://github.com/openshift/library-go/pull/1794


deads2k:
updated for current PR

PR #614, opened 2025-03-18
Implements rand deprecation by removing seed initializer

sebrandon1:
Go 1.20 automatically seeds the top-level generator and deprecated `rand.Seed`.  Read more [here](https://go.dev/blog/randv2#fix.v1).

Similar to: https://github.com/openshift/cert-manager-operator/pull/243


benluddy:
/remove-lifecycle stale
/ok-to-test
/lgtm

Thanks!

PR #630, opened 2025-11-12
CNTRLPLANE-1544: manifests: Use restricted-v3 scc for the operator

tchap:
This effectively enabled user namespaces for the operator.


p0lyn0mial:
@tchap 

`Waiting for status to be reported — Not mergeable. Needs verified label.`

PR #657, opened 2026-02-24
CNTRLPLANE-2619: make checkendpoints use apiserver config

ricardomaraschini:
in order to make the check-endpoints container to respect the global tls config we need to make it point to a config where the tls is present.

the check-endpoints container accepts a --config flag, such flag points to a file containing a OpenShiftAPIServerConfig.


ricardomaraschini:
@gangwgr @kaleemsiddiqu This one also needs proper IPv6 validation.

PR #689, opened 2026-05-14
CNTRLPLANE-3237: Bump library-go and sync encryption tests with APIServerEncryption struct change

gangwgr:
[CNTRLPLANE-3237](https://redhat.atlassian.net/browse/CNTRLPLANE-3237): Bump library-go and sync encryption tests with APIServerEncryption struct change

## Summary by CodeRabbit

* **Chores**
  * Updated OpenShift API, client-go, and library-go dependencies to newer versions

* **Tests**
  * Adjusted KMS encryption end-to-end test setup to use the current encryption provider configuration
  * Aligned encryption rotation and enable/disable tests to the expected encryption provider structure for consistency


_output/golang-versions
```diff
@@ -0,0 +1,3 @@
+1.25
```
ardaguclu: This file shouldn't be here?

_output/named-golang-versions
```diff
@@ -0,0 +1,3 @@
+.ci-operator.yaml: 1.25
```
ardaguclu: This file shouldn't be here?

ardaguclu:
Great. Thank you. That would be awesome, if we can carry those changes to others as well.
/lgtm

PR #765, opened 2026-09-07
OCPBUGS-23746:  operator: suppress brief Available=False blips

tchap:
Add a 10-second inertia for the APIServicesAvailable and
APIServerDeploymentAvailable sub-conditions. This prevents transient
1–4 second gaps — caused by endpoint slices momentarily losing all
addresses when the openshift-apiserver deployment is rolled, or by the
deployment object being briefly unretrievable while the CVO patches it —
from propagating to ClusterOperator.Available=False.

The inertia is scoped to exactly these two sub-conditions (zero default)
so no other Available contributor is affected. Genuine outages that
persist beyond 10 s still surface immediately.

More details are available in a doc linked in a comment on the Jira issue.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved API server availability reporting by preventing brief, transient failures from immediately changing availability status.
  - Availability status now allows a 10-second stabilization period for key API service and deployment conditions.
  - Other availability conditions continue to be reported without this delay.

- **Chores**
  - Updated an underlying OpenShift library dependency.

### DIFF
--- a/pkg/operator/starter.go
+++ b/pkg/operator/starter.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"regexp"
 	"time"
 
 	"k8s.io/apimachinery/pkg/runtime"
@@ -206,6 +207,9 @@ func RunOperator(ctx context.Context, controllerConfig *controllercmd.Controller
 	if infra == nil || infra.Status.ControlPlaneTopology != configv1.SingleReplicaTopologyMode {
 		statusControllerOptions = append(statusControllerOptions, apiservercontrollerset.WithStatusControllerPdbCompatibleHighInertia("APIServer"))
 	}
+	statusControllerOptions = append(statusControllerOptions, func(s *status.StatusSyncer) *status.StatusSyncer {
+		return s.WithAvailableInertia(newAvailableInertia())
+	})
 
 	oasEncryptionStatusProvider, err := encryptionstatusprovider.NewOpenShiftAPIServerEncryptionStatusProvider(operatorConfigClient)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/operator/starter.go
```diff
@@ -541,3 +545,23 @@ func extractOperatorStatus(obj *unstructured.Unstructured, fieldManager string)
...
+		// while the CVO patches it.  Zero default so no other Available
+		// sub-condition is affected.
+		inertiaForCondition("APIServicesAvailable", 10*time.Second),
+		inertiaForCondition("APIServerDeploymentAvailable", 10*time.Second),
+	).Inertia
+}
+
+func inertiaForCondition(cond string, duration time.Duration) status.InertiaCondition {
+	return status.InertiaCondition{
+		ConditionTypeMatcher: regexp.MustCompile("^" + cond + "$"),
```
bertinatto, line 564: if a new `XYZAvailable` condition is added, we'd likely need to add it here as well. Should the regex cover all `Available` conditions?

pkg/operator/starter.go
```diff
@@ -541,3 +545,23 @@ func extractOperatorStatus(obj *unstructured.Unstructured, fieldManager string)
...
+func newAvailableInertia() status.Inertia {
+	return status.MustNewInertia(
+		0,
+		// Suppress brief Available=False blips (1–4 s) that occur when the
+		// openshift-apiserver deployment is rolled: endpoint slices momentarily
+		// lose all addresses, or the deployment object is briefly unretrievable
+		// while the CVO patches it.  Zero default so no other Available
+		// sub-condition is affected.
+		inertiaForCondition("APIServicesAvailable", 10*time.Second),
+		inertiaForCondition("APIServerDeploymentAvailable", 10*time.Second),
```
bertinatto, line 558: have we proof that 10s is what we need/want? Are there CI jobs we can schedule in this PR that would tell us whether the problem would be solved? Or do we need manual testing?

pkg/operator/starter.go
tchap: I try to be as specific as possible not to cover for other potential issues. These are the conditions I researched and I vote for ignoring for 10 seconds.

pkg/operator/starter.go
tchap: This is kinda hard to catch in a single CI run, because it's probabilistic. It happens occasionally. But I did my research and you can check it as a [Google Doc](https://drive.google.com/file/d/1ygjEdnRewhcKd2eC7h7S_AqB6gd-7PQC/view). That's where the 10s interval comes from.

